PR #1, opened 2024-06-26
Add api

ltitanb:
Adds a formal `Signer API` spec and a simple client implementation. To clarify the naming:
- the Signer module implements the Signer API and provides a light abstraction over different signing methods
- Commit Modules call the Signer API to request signatures , etc

Currently the default `SignerClient` can be used, it expects a url endpoint and a jwt for auth, which will need to be passed by the launcher eg via ENV variables or config

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -68,30 +68,53 @@ impl Args {
         match self.cmd {
             Command::Start { config: config_path } => {
                 let config = CommitBoostConfig::from_file(&config_path);
-
-                // TODO: fix with config
-                let pbs_jwt = "MY_PBS_TOKEN";
                 let signer_config = config.signer.expect("missing signer config with modules");
+
+                // start signing server
+                // TODO: generate jwt for each module id
+                let pbs_jwt = "MY_PBS_TOKEN";
+                let jwts = HashMap::from([("PBS_DEFAULT".into(), pbs_jwt.into())]);
                 let signer_address = signer_config.address;
 
+                // Initialize Docker client
+                let docker = bollard::Docker::connect_with_local_defaults()
+                    .expect("Failed to connect to Docker");
+
                 if let Some(modules) = config.modules {
-                    // this mocks the commit boost client starting containers, processes etc
-                    let mut child_handles = Vec::with_capacity(modules.len());
+                    // start signing server
+                    tokio::spawn(SigningService::run(config.chain, signer_config, jwts));
 
                     for module in modules {
-                        let child = std::process::Command::new(module.path)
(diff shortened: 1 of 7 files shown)


crates/common/src/commit/client.rs
```diff
@@ -0,0 +1,80 @@
...
+use super::{error::SignerClientError, request::SignRequest};
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct GetPubkeysResponse {
+    pub consensus: Vec<BlsPublicKey>,
+    pub proxy: Vec<BlsPublicKey>,
+}
+
+const GET_PUBKEYS_PATH: &str = "/signer/v1/pubkeys";
+const REQUEST_SIGNATURE_PATH: &str = "/signer/v1/request_signature";
```
David-Petrov, line 14: I'm not sure at all why the change of the API was necessary and why these new constants have been introduced.
Seeing that we don't have many users yet, I guess changing APIs is fine, but this generally comes with a lot of precaution for me :D

[NIT] In any case, I'd think about unifying the `SignerClient`'s method names with the endpoints. Depends on how much we wanna think about this at this stage, but it's good to keep them corresponding.

I'd change the `pubkeys` endpoint to `get_pubkeys`, so that we have both `get_pubkeys` and `request_signature` starting with a verb :smile:

David-Petrov: Oh, I see, that's what I was missing as context, it's clear now. A second PR is fine 👍

crates/cli/src/lib.rs
```diff
@@ -69,9 +69,12 @@ impl Args {
             Command::Start { config: config_path } => {
                 let config = CommitBoostConfig::from_file(&config_path);
 
-                if let Some(modules) = config.modules {
-                    let signer_config = config.signer.expect("missing signer config with modules");
+                // TODO: fix with config
+                let pbs_jwt = "MY_PBS_TOKEN";
```
David-Petrov, line 73: I'd move this closer to where it's needed (inside the `else` branch).
As a side question to clarify: this line is quasi "generating" the token, right?

crates/cli/src/lib.rs
```diff
@@ -86,7 +89,9 @@ impl Args {
                     }
 
                     // start signing server
-                    tokio::spawn(SigningService::run(config.chain, signer_config));
+                    // TODO: generate jwt for each module id
+                    let jwts = HashMap::new();
```
David-Petrov, line 93: If I understand correctly: `jwts` should contain all the module JWTs **and** the `pbs_jwt`, right?

David-Petrov:
LGTM 👍

PR #3, opened 2024-06-26
feat: launch module from docker image

David-Petrov:
Adds the ability to also launch a module by creating and starting a container by a docker image provided for that module.

The `config.example.toml` is modified to launch an example replica of `da_commit` which has been dockerized by the image `da_commit`, which is not defined in this repo, but is available [here](https://github.com/LimeChain/da-commit-dockerized).

Before you launch CB, build the `da_commit` image locally. Keep in mind that you'll have to modify `Cargo.toml` to point to the accurate current version of CB before you build the image.

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -17,6 +17,11 @@ address = "127.0.0.1:33950"
 key_path = "keys.example.json"
 
 [[modules]]
-id = "DA_COMMIT"
+id = "DA_COMMIT_RAW"
 path = "target/debug/da_commit"
+sleep_secs = 5
+
+[[modules]]
+id = "DA_COMMIT"
+docker_image="da_commit"
 sleep_secs = 5
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -1,7 +1,7 @@
 use std::process::Stdio;
 
 use cb_common::{
-    config::{CommitBoostConfig, CONFIG_PATH_ENV, MODULE_ID_ENV},
+    config::{CommitBoostConfig, ModuleSource, CONFIG_PATH_ENV, MODULE_ID_ENV},
     utils::print_logo,
 };
 use cb_crypto::service::SigningService;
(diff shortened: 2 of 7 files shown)


crates/common/src/config.rs
```diff
@@ -113,13 +113,57 @@ const fn default_u256() -> U256 {
...
 pub struct ModuleConfig<T = ()> {
     pub id: String,
-    pub path: String,
+    #[serde(flatten)]
+    pub source: ModuleSource,
     #[serde(flatten)]
     pub extra: T,
 }
 
+impl<'de, T> Deserialize<'de> for ModuleConfig<T>
```
David-Petrov, line 133: That's the quick approach I took to scratch my itch for modelling a strict config structure: exactly one of `path` or `docker_image` must be specified in the `config.example.toml` for a given module - marking the module as either launchable from a raw binary (like it used to be) OR from a docker image (the new addition).

We could choose another approach for modelling the config (i.e. an additional tag specifying the variant, and then the necessary source data (path or docker image id); or similar...)

crates/cli/src/lib.rs
```diff
@@ -69,24 +69,53 @@ impl Args {
...
+                        match module.source {
+                            ModuleSource::DockerImageId(docker_image) => {
+                                let config = bollard::container::Config {
+                                    image: Some(docker_image.clone()),
+                                    host_config: Some(bollard::secret::HostConfig {
+                                        binds: {
+                                            let full_config_path = std::fs::canonicalize(&config_path).unwrap().to_string_lossy().to_string();
+                                            Some(vec![format!("{}:{}", full_config_path, "/config.toml")])
+                                        },
+                                        network_mode: Some(String::from("host")), // Use the host network
```
David-Petrov, line 93: It's worth noting here that I'm making substantial use of the [host network mode](https://docs.docker.com/network/drivers/host/) in order to share the networking namespace of the host with the module itself. 
This is sacrificing sandboxing for easier configuration. 

*NOTE* This is a [beta feature for Docker desktop](https://docs.docker.com/network/drivers/host/#docker-desktop) (if you're using that), and must be enabled explicitly.

Might not be the best long-term decision, but for simplicity's sake it's closer to native; just like the alternative (previous) approach of launching modules as subprocesses.

config.example.toml
```diff
@@ -17,6 +17,11 @@ address = "127.0.0.1:33950"
...
 
 [[modules]]
-id = "DA_COMMIT"
+id = "DA_COMMIT_RAW"
 path = "target/debug/da_commit"
 sleep_secs = 5
+
+[[modules]]
+id = "DA_COMMIT"
+docker_image="da_commit"
```
David-Petrov, line 26: That's the example image that must be available in the local registry before starting commit-boost.

crates/cli/src/lib.rs
```diff
@@ -69,24 +69,53 @@ impl Args {
             Command::Start { config: config_path } => {
                 let config = CommitBoostConfig::from_file(&config_path);
 
+                // Initialize Docker client
+                let docker = bollard::Docker::connect_with_local_defaults().expect("Failed to connect to Docker");
```
David-Petrov, line 73: Here, we currently expect that CB and the docker engine run on the same machine. 

This is, however, the way we connect to the docker daemon is modifiable if more complex needs arise: [ref](https://github.com/fussybeaver/bollard/tree/master?tab=readme-ov-file#connecting-with-the-docker-daemon).

David-Petrov:
There are currently no issues introduced with the support of binary sourced modules. In fact, the current `config.example.toml` showcases simultaneous inclusion of on copy of `da_commit`, sourced as a binary, and one copy of `da_commit` sourced from a docker image.

You might be right, though, that we might run into difficulties at some point... so I guess we can then remove the binary option.

Does that change perspectives or do we stick to removing binary sources right away?

PR #37, opened 2024-07-24
expand metrics

### LINKED ISSUE #35: Expand PBS metrics
We currently collect some metrics in the [PBS module](https://github.com/Commit-Boost/commit-boost-client/blob/main/crates/pbs/src/metrics.rs). We should expand those to have more granular collection

ltitanb:
fix  #35

- add more metrics for PBS module based on feedback from node operators


David-Petrov (approved):
Looks good to me! 👍

Only thing that's getting a bit difficult for me is keeping track of all the string constants that are piling up. Perhaps we could consider slowly migrating them into structs/enums to group them more granularly (improve namespacing and semantics), or similar?
Just a little nitty consideration I'm sharing, nothing to worry about necessarily. 😄

PR #40, opened 2024-07-29
timing games

### LINKED ISSUE #23: Add support for timing games
We should add support in the default PBS module for timing games such as making multiple `get_header` requests to relays, and/or sending them delayed. These should be off by default and configurable by node operators

ltitanb:
fix #23 

This PR adds support for timing games, ie. allows a node operator to fine tune how to send `get_header` requests to relays. This is configured via two new per-relay parameters:
- `target_first_request_ms`: which sets how late in the slot to send the first header request
- `frequency_get_header_ms`: which sets how frequently to send header requests after the first

and a global param:
- `late_in_slot_time_ms`: which sets how late in the slot is "late", acting as a hard upper bound for when to return headers to the CL

The requests are still bound by `timeout_get_header_ms` which is ~what the full `get_header` request can take at most from the CL perspective (plus some small overhead from the module itself).

The logic is as follows:
- if the request from the CL comes before `target_first_request_ms`, wait until the target time
- after that send requests every `frequency_get_header_ms` with decreasing timeouts
- if the request comes later into the slot, then reduce the max timeout accordingly

**Examples**
Assuming: `timeout_get_header_ms = 950`, `frequency_get_header_ms = 300`, `target_first_request_ms = 200`, `late_in_slot_time_ms = 2000`

1) CL request comes at 100ms in the slot (max timeout 1050ms in the slot), then:
- sleep for 100ms
- send request at 200ms with 850ms timeout
- send request at 500ms with 550ms timeout
- send request at 800ms with 250ms timeout

2) CL request comes at 1500ms in the slot (max timeout 2000ms in the slot), then:
- send request at 1500ms with 500ms timeout
- send request at 1800ms with 200ms timeout

3) CL request comes 2500ms in the slot then:
- return 204 and force local build

### DIFF
--- a/crates/common/src/pbs/config.rs
+++ b/crates/common/src/pbs/config.rs
@@ -21,6 +21,7 @@ pub struct RelayConfig {
     /// Optional headers to send with each request
     pub headers: Option<HashMap<String, String>>,
     /// Whether to enable timing games
+    #[serde(default = "default_bool::<false>")]
     pub enable_timing_games: bool,
     /// Target time in slot when to send the first header request
     pub target_first_request_ms: Option<u64>,
--- a/crates/pbs/src/mev_boost/get_header.rs
+++ b/crates/pbs/src/mev_boost/get_header.rs
@@ -14,7 +14,7 @@ use cb_common::{
 use futures::future::join_all;
 use reqwest::{header::USER_AGENT, StatusCode};
 use tokio::time::sleep;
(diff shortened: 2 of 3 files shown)


cwhcheng:
Hi! If a relay intentionally delays the return of the get_header response to the max timeout value, does it make any difference whether the validators try to send multiple get_header request within the timeout?

PR #44, opened 2024-07-31
Feat/prometheus persist

mikirov:
Programatically adds a docker volume on to the docker-compose to be used for storing prometheus data


docs/index.html
```diff
@@ -0,0 +1 @@
+# Test
```
Daniel-K-Ivanov, line 1: is this change necessary?

Daniel-K-Ivanov (changes requested):

Daniel-K-Ivanov (commented):
There are `main` branch conflicts

PR #46, opened 2024-08-01
builder events module

### LINKED ISSUE #26: Add support for `BuilderApi` events
We need to add support for modules to subscribe to builder Api events

https://github.com/Commit-Boost/commit-boost-client/blob/e6ad292ee43063c16e397e58a9a152482306d16f/crates/pbs/src/boost.rs#L58-L68

We should also add some examples such as sending a notification to telegram / discord  if a payload is missed

ltitanb:
fix #26

Add support for PBS buiilder event modules. 
We now support two kind of modules: "commit" , and "events". 
The PBS module now makes HTTP calls to the "events" modules

### DIFF
--- a/crates/common/src/config/module.rs
+++ b/crates/common/src/config/module.rs
@@ -1,4 +1,4 @@
-use eyre::{eyre, ContextCompat, Result};
+use eyre::{ContextCompat, Result};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use toml::Table;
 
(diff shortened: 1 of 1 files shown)


crates/common/src/config/module.rs
```diff
@@ -0,0 +1,174 @@
...
+    struct StubConfig<U> {
+        chain: Chain,
+        modules: Vec<ThisModule<U>>,
+    }
+
+    // load module config including the extra data (if any)
+    let cb_config: StubConfig<T> = load_file_from_env(CB_CONFIG_ENV);
+
+    // find all matching modules config
+    let matches: Vec<ThisModuleConfig<T>> = cb_config
```
fbrv, line 153: bit more readable
```rust
let matches: Vec<ThisModuleConfig<T>> = cb_config
    .modules
    .into_iter()
    .filter_map(|m| match m {
        ThisModule::Target(config) => Some(config),
        _ => None,
    })
    .collect();
```

crates/common/src/pbs/event.rs
```diff
@@ -0,0 +1,121 @@
...
+        })
+    }
+
+    pub fn publish(&self, event: BuilderEvent) {
+        for endpoint in self.endpoints.clone() {
+            let client = self.client.clone();
+            let event = event.clone();
+
+            tokio::spawn(async move {
+                trace!("Publishing event to {}", endpoint);
```
fbrv, line 64: publishing refers more to a pubsub scenario, here seems more "sending event to blah"

crates/common/src/pbs/event.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        let router = axum::Router::new()
+            .route(BUILDER_EVENTS_PATH, post(handle_builder_event::<T>))
+            .with_state(self.processor);
+        let address = SocketAddr::from(([0, 0, 0, 0], self.port));
+        let listener = TcpListener::bind(&address).await?;
+
+        axum::serve(listener, router).await?;
+
+        bail!("Builder events stopped")
```
fbrv, line 104: the error seems irrecoverable, maybe panic here?

fbrv (approved):

PR #47, opened 2024-08-01
Feat/logs

mikirov:
Adds a new binary and docker service that creates a new log file each day and retains log files for specified period (30 days by default).


fbrv:
Hey @mikirov I think that the log rotation is a bit convoluted, I quickly opened a draft pr, #48 where you can see the usage of tracing_appender crate

mikirov:
> Hey @mikirov I think that the log rotation is a bit convoluted, I quickly opened a draft pr, #48 where you can see the usage of tracing_appender crate

I like the simplicity and elegance of this approach. Let me close this PR and lets iterate on the one you opened

mikirov:
Will continue with the approach in #48

PR #66, opened 2024-08-06
add commit boost to UA

### LINKED ISSUE #65: Add version to UserAgent header
We should add the Commit Boost version to the `UserAgent` header. Similar to
https://github.com/flashbots/mev-boost/blob/c2a5d005161e09d59aa4d42f936d60e2aa65a568/server/utils.go#L67

ltitanb:
fix #65

### DIFF
--- a/crates/common/src/utils.rs
+++ b/crates/common/src/utils.rs
@@ -221,7 +221,8 @@ pub fn random_jwt() -> String {
     rand::thread_rng().sample_iter(&Alphanumeric).take(32).map(char::from).collect()
 }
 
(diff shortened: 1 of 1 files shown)


crates/common/src/utils.rs
```diff
@@ -220,8 +222,15 @@ pub fn random_jwt() -> String {
 }
 
 /// Extracts the user agent from the request headers
```
fbrv, line 224: should we update the comment?

fbrv (approved):

PR #72, opened 2024-08-08
simplify builider api generics

ltitanb:
- Simplify the `BuilderApi` and `BuilderApiState` traits
- Add a working example on how to use the Api in `examples/status_api`

fix issues from #69


fbrv (approved):
gtg, we should a bit more descriptive with generics name. Ie, 
```rust
struct PbsState<S> {
  data: S
}
```
could be renamed as 
```rust
puct struct PbsState<EData> {
  data: EData,
}
```
In order to make the code a easier to read in case someone miss bits of documentation

David-Petrov (approved):
Everything looks fine!

Regarding what Fabio said about generic type parameters' names, I'm also a fan of more descriptive (whole-word even) type names, but this can easily become cumbersome, too. For me, right now a satisfactory solution would be to just describe type parameters in the rustdoc of the struct/trait that introduces them.

PR #73, opened 2024-08-09
Add relay monitors

### LINKED ISSUE #29: Add relay monitor support
The [MEV Boost](https://github.com/flashbots/mev-boost/blob/develop/server/service.go#L64) client supports relay monitors which are essentially remote metrics collectors. We should add similar support from the default PBS module. This can likely be implemented using builderAPI events #26

ltitanb:
fix #29 

- Add support for relay monitor, similar to https://github.com/flashbots/mev-boost/blob/1fed7bb5138d37ecdfb204895e0c5a2179d735f9/server/service.go#L210-L224
- Change urls from `String` to `Url`, this should avoid some annoying bugs like not being able to parse a trailing slash in the url
- Also added a `validate()` function in the configs, not used for now but we should add some runtime validation which can be caught at `init`

### DIFF
--- a/crates/common/src/config/pbs.rs
+++ b/crates/common/src/config/pbs.rs
@@ -3,7 +3,7 @@
 use std::{collections::HashMap, sync::Arc};
 
 use alloy::primitives::U256;
-use eyre::{Context, Result};
+use eyre::Result;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use url::Url;
 
@@ -16,7 +16,7 @@ use crate::{
     utils::{as_eth_str, default_bool, default_u256, default_u64},
 };
 
-#[derive(Debug, Clone, Default, Deserialize, Serialize)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct RelayConfig {
     /// Relay ID, if missing will default to the URL hostname from the entry
     pub id: Option<String>,
@@ -34,7 +34,7 @@ pub struct RelayConfig {
     pub frequency_get_header_ms: Option<u64>,
 }
 
-#[derive(Debug, Clone, Default, Deserialize, Serialize)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct PbsConfig {
     /// Port to receive BuilderAPI calls from beacon node
     pub port: u16,
@@ -58,7 +58,7 @@ pub struct PbsConfig {
     pub min_bid_wei: U256,
     /// List of relay monitor urls in the form of scheme://host
     #[serde(default)]
-    pub relay_monitors: Vec<String>,
+    pub relay_monitors: Vec<Url>,
     /// How late in the slot we consider to be "late"
     #[serde(default = "default_u64::<LATE_IN_SLOT_TIME_MS>")]
     pub late_in_slot_time_ms: u64,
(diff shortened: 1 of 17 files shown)


crates/common/src/config/pbs.rs
```diff
@@ -55,13 +56,23 @@ pub struct PbsConfig {
...
+    pub relay_monitors: Vec<Url>,
     /// How late in the slot we consider to be "late"
     #[serde(default = "default_u64::<LATE_IN_SLOT_TIME_MS>")]
     pub late_in_slot_time_ms: u64,
 }
 
+impl PbsConfig {
+    /// Validate PBS config parameters
+    pub fn validate(&self) -> Result<()> {
+        Ok(())
```
fbrv, line 70: no logic is needed to validate the pbs config?

crates/common/src/pbs/relay.rs
```diff
@@ -90,46 +90,87 @@ impl RelayClient {
     }
 
     // URL builders
-    pub fn get_url(&self, path: &str) -> String {
-        format!("{}{path}", &self.config.entry.url)
+    pub fn get_url(&self, path: &str) -> Result<Url, PbsError> {
```
fbrv, line 93: why not using direclty eyre::Result

crates/common/src/pbs/relay.rs
```diff
@@ -90,46 +90,87 @@ impl RelayClient {
...
+        self.config.entry.url.join(path).map_err(PbsError::UrlParsing)
+    }
+    pub fn builder_api_url(&self, path: &str) -> Result<Url, PbsError> {
```
fbrv, line 96: as above

crates/common/src/pbs/error.rs
fbrv, line 1: mmh, from the pr inspection it seems that PbsError is used only in pbs crate, I would keep where it was. If we are using it outside than makes sense to have it in common

crates/pbs/src/routes/register_validator.rs
```diff
@@ -26,6 +32,13 @@ pub async fn handle_register_validator<S: BuilderApiState, A: BuilderApi<S>>(
 
     info!(ua, num_registrations = registrations.len());
 
+    if state.has_monitors() {
+        // send registrations to monitors
+        for relay_monitor in state.pbs_config().relay_monitors.clone() {
+            tokio::spawn(send_relay_monitor_registrations(registrations.clone(), relay_monitor));
```
fbrv, line 38: if a task if failing we are not managing the error.
We could do something like
```rust
async fn send_relay_monitor_registrations(
    registrations: Vec<ValidatorRegistration>,
    relay_monitor_url: Url,
) -> eyre::Result<()> {...}
```
And then handling the error like

```rust
let handles: Vec<_> = state.pbs_config().relay_monitors.clone().into_iter().map(|relay_monitor| {
    let registrations_clone = registrations.clone();
    tokio::spawn(async move {
        send_relay_monitor_registrations(registrations_clone, relay_monitor).await
    })
}).collect();

let results: Result<Vec<_>, _> = try_join!(handles.into_iter().map(|h| h));
match results {
    Ok(_) => println!("All tasks completed successfully."),
    Err(e) => eprintln!("A task failed: {:?}", e),
}
```

fbrv (approved):

David-Petrov (approved):
LGTM! 👍

Issue #36, opened 2024-07-24
Tidy up ssz types


ltitanb:
We're currently using the `tree_hash` crate, and some hacky workaround since the crate doesn't support alloy yet, eg:

https://github.com/Commit-Boost/commit-boost-client/blob/cadefca902a8308b58b03976777d6a4a75ad8cd2/crates/pbs/src/types/execution_payload.rs#L85-L86

We should remove the workarounds once [this PR](https://github.com/sigp/tree_hash/pull/16) is merged

David-Petrov:
The referred PR was merged today, so this issue is ready to pick up.

PR #87, opened 2024-08-14
feat(signer): add ECDSA proxy keys

### LINKED ISSUE #20: Add support for ECDSA proxy delegation
Currently proxy keys can only be Bls. We should add the option to delegate to a ECDSA key which has better on-chain support

David-Petrov:
# Quick summary:
The first two commits are the main content of this PR, so changes are roughly two-fold:
- refactor of existing infrastructure to prepare support for different signature schemes (other than BLS) for proxy signatures;
- actually implementing ECDSA as a second signature scheme, whilst further refining the multiple signature infrastructure in the process.

The rest of the commits are smaller decisions and easier to follow.

# Update 1
From eba97b0 below, the PR takes an alternative approach to implementing ECDSA proxies where we strive to flatten the flow by losing the generics and making the client SDK more type safe.

# TODOs:
- [x] Take care of local inline TODOs.
- [x] Better serialization to request/response structs for a proper REST API
- [x] Update Signer API docs (with examples).
- [x] Add tests for the ECDSA keys.

---

fixes #20


crates/common/src/commit/request.rs
```diff
@@ -32,19 +43,28 @@ impl SignedProxyDelegation {
     }
 }
 
+impl fmt::Display for SignedProxyDelegation {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}\nSignature: {}", self.message, self.signature)
+    }
+}
+
+// TODO(David): Consider splitting `SignRequest` into two: ConsensusSignRequest
```
David-Petrov, line 52: Split this up.

crates/common/src/commit/client.rs
```diff
@@ -69,7 +70,7 @@ impl SignerClient {
     pub async fn request_signature(
```
David-Petrov, line 70: Split this client SDK method, too, in order to provide a couple of well-typed methods (with proper return types) instead of one, which is handling opaque byte vectors.

crates/common/src/signer/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+use tree_hash::TreeHash;
+
+pub mod schemes;
+#[allow(clippy::module_inception)]
+mod signer;
+
+pub use schemes::{bls::BlsSecretKey, ecdsa::EcdsaSecretKey};
+pub use signer::{ConsensusSigner, Signer};
+
+pub type Pubkey<T> = <T as SecretKey>::PublicKey;
```
David-Petrov, line 18: We'd like to avoid this type alias and stick to concrete types (opinionated is not that bad in our case).

crates/common/src/signer/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+
+pub trait SecretKey {
```
David-Petrov, line 20: Try getting rid of this trait and stick to the enums only (abstraction of the signature scheme with this trait deemed not worth it for now where we have only two variants).

crates/common/src/signer/schemes/ecdsa.rs
```diff
@@ -0,0 +1,185 @@
...
+
+#[derive(Debug, Clone, Copy, From, Into, Serialize, Deserialize, PartialEq, Eq, Deref, Hash)]
+pub struct EcdsaPublicKey {
+    encoded: CompressedPublicKey,
+}
+
+impl EcdsaPublicKey {
+    /// Size of the public key in bytes. We store the SEC1 encoded affine point
+    /// compressed, thus 33 bytes.
+    const SIZE: usize = 33;
```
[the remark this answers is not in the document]
David-Petrov: Resolved after a bit clarification regarding public key encoding. We stick to the current state.

crates/signer/src/manager.rs
```diff
@@ -16,125 +23,243 @@ use crate::error::SignerModuleError;
...
+
+#[derive(Clone, Deref)]
+pub struct EcdsaProxySigner {
+    #[deref]
+    signer: EcdsaSigner,
+    delegation: SignedProxyDelegationEcdsa,
+}
+
+#[derive(From)]
+pub enum GenericProxySigner {
```
David-Petrov, line 41: I'd say after losing the generics, the enums `GenericPubkey`, `GenericProxySigner` and the sort around `manager.rs`, remain kind of awkwardly in the middle and don't make much sense. If we decide we're satisfied with splitting up pretty much the entire signer module signing logic into two completely separate flows (one for BLS, one for ECDSA), then we can get rid of these uniting enums, too.

Seems to me like the main purpose they're currently serving is a not so grandiose reduction of duplication within `manager.rs`, which could be a cost worth paying... I'm not sure about the overall split, though.

crates/signer/src/manager.rs
```diff
@@ -16,125 +23,183 @@ use crate::error::SignerModuleError;
...
+                self.proxy_signers.bls_signers.get(bls_pk).map(|x| x.delegation).map(Into::into)
+            }
+            GenericPubkey::Ecdsa(ecdsa_pk) => {
+                self.proxy_signers.ecdsa_signers.get(ecdsa_pk).map(|x| x.delegation).map(Into::into)
+            }
+        };
+
+        delegation.ok_or(SignerModuleError::UnknownProxySigner(pubkey.as_ref().to_vec()))
     }
 }
```
David-Petrov, line 198: That's the generic implementation of `get_delegation`. If we deem this unwanted as too complex, then we can either:
1) [the simplest] get rid of this method as it's still unused anywhere
2) go the other way and duplicate: `get_delegation_ecdsa` and `get_delegation_bls`
3) go full circle and make the proxy signers, the signers, etc. also generic (this time with a trait bound to the public key instead of the secret keys), but this is going to infest the signer module with generics overall :D

crates/signer/src/manager.rs
```diff
@@ -180,20 +183,38 @@ impl SigningManager {
...
+    }
+}
 
-        delegation.ok_or(SignerModuleError::UnknownProxySigner(pubkey.as_ref().to_vec()))
+impl GetDelegation<EcdsaPublicKey> for SigningManager {
+    fn get_delegation(
+        &self,
+        pubkey: &EcdsaPublicKey,
+    ) -> Option<SignedProxyDelegation<EcdsaPublicKey>> {
+        self.proxy_signers.ecdsa_signers.get(pubkey).map(|x| x.delegation)
```
David-Petrov, line 217: That's the generic implementation of `get_delegation`. 
If we deem this unwanted as too complex, then we can either:
1) [the simplest] get rid of this method as it's still unused anywhere
2) go the other way and duplicate two methods: `get_delegation_ecdsa` and `get_delegation_bls`
3) go full circle and make the proxy signers, the signers, etc. also generic (this time with a trait bound to the public key instead of the secret keys), but this is going to infest the signer module with generics overall again. :D

crates/common/src/commit/request.rs
David-Petrov: Resolved in [eba97b0](https://github.com/Commit-Boost/commit-boost-client/pull/87/commits/eba97b0ca03b140bbfdc9f4b7b8ad4adf382776b)

crates/common/src/signer/mod.rs
David-Petrov: Resolved in [eba97b0](https://github.com/Commit-Boost/commit-boost-client/pull/87/commits/eba97b0ca03b140bbfdc9f4b7b8ad4adf382776b)

crates/common/src/signature.rs
```diff
@@ -88,25 +50,25 @@ pub fn verify_signed_builder_message<T: TreeHash>(
     let domain = chain.builder_domain();
     let signing_root = compute_signing_root(msg.tree_hash_root().0, domain);
 
-    verify_signature(pubkey, &signing_root, signature)
+    pubkey.verify_signature(&signing_root, signature)
 }
 
-pub fn sign_builder_message(
+pub fn sign_builder_message<T: SecretKey>(
```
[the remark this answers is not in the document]
David-Petrov: Resolved in [eba97b0](https://github.com/Commit-Boost/commit-boost-client/pull/87/commits/eba97b0ca03b140bbfdc9f4b7b8ad4adf382776b)

crates/common/src/commit/client.rs
David-Petrov: Resolved in [eba97b0](https://github.com/Commit-Boost/commit-boost-client/pull/87/commits/eba97b0ca03b140bbfdc9f4b7b8ad4adf382776b)

crates/common/src/commit/client.rs
```diff
@@ -83,17 +91,58 @@ impl SignerClient {
...
+    }
+
+    pub async fn request_consensus_signature(
+        &self,
+        request: SignConsensusRequest,
+    ) -> Result<BlsSignature, SignerClientError> {
+        let request = SignRequest::Consensus(request);
+        let raw_signature = self.request_signature(&request).await?;
+
+        let signature = BlsSignature::from_slice(&raw_signature);
```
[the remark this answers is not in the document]
David-Petrov: It is fallible, it just panics.
This is the place where a deserialization error is not expected as a normal runtime exception; it's a rather ugly runtime assertion that we know the underlying signature will be the correct type (i.e. bls signature) and everything else is a development mistake.
We can always decide we prefer to propagate this as an err value, but it won't be something the client will be able to recover from (by requesting again or similar), but I can agree propagating the error will be safer.

crates/common/src/commit/client.rs
```diff
@@ -83,17 +91,58 @@ impl SignerClient {
...
+        self.request_signature(&request).await
+    }
+
+    pub async fn request_proxy_ecdsa_signature(
+        &self,
+        request: SignProxyEcdsaRequest,
+    ) -> Result<EcdsaSignature, SignerClientError> {
+        let raw_signature = self.request_proxy_signature(request.into()).await?;
+        let signature = EcdsaSignature::try_from(raw_signature.as_ref())
+            .expect("requested signature should be ECDSA");
```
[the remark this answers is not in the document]
David-Petrov: Same as above; will do.

Cargo.toml
```diff
@@ -59,9 +59,12 @@ prometheus = "0.13.4"
 
 # crypto
 blst = "0.3.11"
-tree_hash = "0.5"
-tree_hash_derive = "0.5"
+tree_hash = "0.6.0"
+tree_hash_derive = "0.6.0"
```
David-Petrov, line 63: Double check if updating to this newer version is okay (breaking changes).

crates/common/src/signer/schemes/bls.rs
```diff
@@ -0,0 +1,105 @@
...
+
+// std traits
+#[derive(Debug, Clone, Copy, LowerHex, Display, PartialEq, Eq, Hash)]
+// serde, ssz, tree_hash
+#[derive(Serialize, Deserialize, Encode, Decode, TreeHash)]
+#[ssz(struct_behaviour = "transparent")]
+// derive_more
+#[derive(Deref, From, Into)]
+pub struct BlsPublicKey {
+    inner: BlsPublicKeyInner,
```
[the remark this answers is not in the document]
David-Petrov: I had a quick thought about what we can do regarding opening an issue upstream, but I decided for a wrapper type as a common solution. Now that you propose it, I will have a deeper look into how easy this would be (or maybe if the trait is intentionally not implemented); and eventually open up an issue.

crates/common/src/commit/client.rs
```diff
@@ -83,17 +91,58 @@ impl SignerClient {
             });
         }
 
-        let signature: BlsSignature = serde_json::from_slice(&response_bytes)?;
+        let signature: Vec<u8> = serde_json::from_slice(&response_bytes)?;
```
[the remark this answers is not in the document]
David-Petrov: This is only for quick type unification inside the client SDK, so I'd say it's whatever.

However, the REST API does currently include a lot of not-so-nicely serialised stuff whilst we're experimenting with different approaches, so if that's your concern, it will have to be polished up at the end, sure. Added as a PR task.

crates/signer/src/manager.rs
David-Petrov: Agreed. Refactored in [7ddcd3a](https://github.com/Commit-Boost/commit-boost-client/pull/87/commits/7ddcd3a67764367d26efac87f59509f3bc437505)

crates/common/src/signer/schemes/bls.rs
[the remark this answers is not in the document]
David-Petrov: Another possible argument in favour of a wrapper type would be SDK unification, because we currently have a custom type for the `EcdsaPublicKey` (since `k256::ecdsa` doesn't export a convenient type to represent an encoded point as we need it; like alloy does for `BlsPublicKey`). 
So, currently, module authors would include the `EcdsaPublicKey` from our common crate, but they'll have to import the proper version of `BlsPublicKey` directly from alloy (and care about matching versions). So I'd say this makes sense for us to reexport as a custom wrapper type in that regard, although it does, indeed, add a bit of complexity. 🤔

crates/signer/src/manager.rs
David-Petrov: `SignedProxyDelegation` already generic in [a7d53a1](https://github.com/Commit-Boost/commit-boost-client/pull/87/commits/a7d53a15ed13942d361008f28e0f9886516f4083)

David-Petrov: `SignProxyRequest` also generic in the public key as of [780299e](https://github.com/Commit-Boost/commit-boost-client/pull/87/commits/780299e6b529a96ff9582494868cab559d1b87dc)

This seems like a nice place to stop with the generics - client SDK simplified without any sprawling of generics deeper into the signer module.

crates/common/src/commit/client.rs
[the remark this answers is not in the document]
David-Petrov: Also fixed by [b8f228b](https://github.com/Commit-Boost/commit-boost-client/pull/87/commits/b8f228b630f1865439959b166b5a87a5a23bc0b9)

crates/common/src/signer/schemes/bls.rs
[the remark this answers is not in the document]
David-Petrov: Opened up the issue [here](https://github.com/sigp/tree_hash/issues/22). 

Implementing it locally wasn't a problem, but then we'll have some versioning issues after it gets merged, which tie into [this task](https://github.com/Commit-Boost/commit-boost-client/issues/36).

Leaving the wrapper type as a temporary solution for now.

crates/common/src/commit/client.rs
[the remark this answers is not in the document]
David-Petrov: Intermediate `Vec<u8>` deserialization removed in [b8f228b](https://github.com/Commit-Boost/commit-boost-client/pull/87/commits/b8f228b630f1865439959b166b5a87a5a23bc0b9). 

Rest API hex encoding of request/response structs fixed in [dd864ec](https://github.com/Commit-Boost/commit-boost-client/pull/87/commits/dd864ec2d2fe7e2e3a53003c511431a51388bb1e)

crates/common/src/commit/client.rs
```diff
@@ -60,16 +65,13 @@ impl SignerClient {
             });
         }
 
-        let parsed_response: GetPubkeysResponse = serde_json::from_slice(&response_bytes)?;
+        let parsed_response = serde_json::from_slice(&response_bytes)?;
 
         Ok(parsed_response)
```
fbrv, line 70:
```rust
pub async fn get_pubkeys(&self) -> Result<GetPubkeysResponse, SignerClientError> {
        let res = self.client.get(&format!("{}{}", self.url, GET_PUBKEYS_PATH)).send().await?;

        if !res.status().is_success() {
            return Err(SignerClientError::FailedRequest {
                status: res.status().as_u16(),
                error_msg: String::from_utf8_lossy(&res.bytes().await?).into_owned(),
            });
        }

        Ok(serde_json::from_slice(&res.bytes().await?)?)
    }
```

crates/common/src/commit/client.rs
```diff
@@ -83,17 +85,47 @@ impl SignerClient {
             });
         }
 
-        let signature: BlsSignature = serde_json::from_slice(&response_bytes)?;
+        Ok(response_bytes)
+    }
+
+    pub async fn request_consensus_signature(
```
fbrv, line 91:
```rust
    pub async fn request_consensus_signature(
        &self,
        request: SignConsensusRequest,
    ) -> Result<BlsSignature, SignerClientError> {
        let raw_signature = self.request_signature(&SignRequest::Consensus(request)).await?;
        Ok(serde_json::from_slice(&raw_signature)?)
    }
```

fbrv: actually better
```rust
    async fn request_signature_generic<T: for<'de> serde::Deserialize<'de>>(
        &self,
        request: &SignRequest,
    ) -> Result<T, SignerClientError> {
        let raw_signature = self.request_signature(request).await?;
        Ok(serde_json::from_slice(&raw_signature)?)
    }

    pub async fn request_consensus_signature(
        &self,
        request: SignConsensusRequest,
    ) -> Result<BlsSignature, SignerClientError> {
        self.request_signature_generic(&SignRequest::Consensus(request)).await
    }

    pub async fn request_proxy_ecdsa_signature(
        &self,
        request: SignProxyRequest<EcdsaPublicKey>,
    ) -> Result<EcdsaSignature, SignerClientError> {
        self.request_signature_generic(&request.into()).await
    }

    pub async fn request_proxy_bls_signature(
        &self,
        request: SignProxyRequest<BlsPublicKey>,
    ) -> Result<BlsSignature, SignerClientError> {
        self.request_signature_generic(&request.into()).await
    }
```

crates/common/src/commit/request.rs
```diff
@@ -32,24 +62,60 @@ impl SignedProxyDelegation {
...
     pub pubkey: BlsPublicKey,
-    pub is_proxy: bool,
+    #[serde(with = "alloy::hex::serde")]
     pub object_root: [u8; 32],
 }
 
-impl SignRequest {
-    pub fn new(pubkey: BlsPublicKey, is_proxy: bool, object_root: [u8; 32]) -> SignRequest {
-        Self { pubkey, is_proxy, object_root }
+impl SignConsensusRequest {
```
fbrv, line 87: I wonder if we can do something like:

```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct SignRequestType<T: PublicKey> {
    pub pubkey: T,
    #[serde(with = "alloy::hex::serde")]
    pub object_root: [u8; 32],
}

impl<T: PublicKey> SignRequestType<T> {
    pub fn new(pubkey: T, object_root: [u8; 32]) -> Self {
        Self { pubkey, object_root }
    }

    pub fn builder(pubkey: T) -> Self {
        Self::new(pubkey, [0; 32])
    }

    pub fn with_root(self, object_root: [u8; 32]) -> Self {
        Self { object_root, ..self }
    }

    pub fn with_msg(self, msg: &impl TreeHash) -> Self {
        self.with_root(msg.tree_hash_root().0)
    }
}

// Example usage for BlsPublicKey:
pub type SignConsensusRequest = SignRequestType<BlsPublicKey>;

// Example usage for other types of PublicKey:
pub type SignProxyRequest<T> = SignRequestType<T>;
```
to reduce code duplication

crates/common/src/signer/schemes/bls.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+impl AsRef<[u8]> for BlsPublicKey {
+    fn as_ref(&self) -> &[u8] {
+        self.as_slice()
+    }
+}
+
+#[derive(Clone)]
+pub enum BlsSigner {
```
fbrv, line 37: not entirely sure, why an enum with only one variant?

crates/common/src/signer/schemes/ecdsa.rs
```diff
@@ -0,0 +1,229 @@
...
+        // Unnecessary copying into a `FixedVector` just for its `tree_hash_root`
+        // implementation.  If this becomes a performance issue, we could use
+        // `ssz_types::tree_hash::vec_tree_hash_root`,  which is unfortunately
+        // not public.
+        let vec = self.encoded.to_vec();
+        FixedVector::<u8, U33>::from(vec).tree_hash_root()
+    }
+}
+
+impl ssz::Encode for EcdsaPublicKey {
```
fbrv, line 64: nit: leave as is but rust compiler in general is smart enough to inline the one liner function that return value

crates/common/src/signer/schemes/bls.rs
David-Petrov: That's inherited from what was previously set up in the old bare `Signer`; the other variant would be a remote signer (to be implemented in future). That's also the reason why some of the methods are async (although they're synchronous for the local variant).

crates/common/src/commit/request.rs
David-Petrov: I actually had exactly this going on, but then decided to explicitly differentiate the request types for a consensus request and for a bls proxy request as we want the user-facing API to be as straightforward as possible. A type alias can be considered helpful enough, but we lose the type guarantee we now have after the use has explicitly constructed a request type called `SignConsensusRequest` 😄 So with the suggested type aliasing, technically a `SignProxyRequest<BlsPublicKey>` could still be passed as an argument to `request_consensus_signature`.

With this in mind, I'd still say the code duplication is worth reducing, so if you deem the little drawback outlined above is worth sacrificing, I'm on board with that 👍

crates/common/src/commit/client.rs
David-Petrov: Applied in [4e652d9](https://github.com/Commit-Boost/commit-boost-client/pull/87/commits/4e652d97d33ce7dc3051db072a0f0298878ee315)

crates/common/src/commit/request.rs
fbrv: It's a valid point, let's keep for now, eventually we will change afterwards

crates/common/src/signer/schemes/bls.rs
fbrv: ok, fair enough. In general I prefer to simplify code where possible but remote seems something we will implement sooner rather than later so let's keep as is

Cargo.toml
David-Petrov: The necessity for this upgrade seems to have been eliminated during refactoring. Reverting as not immediately necessary. 👍

PR #106, opened 2024-09-04
add custom signing domain

### LINKED ISSUE #70: Support a commit boost signing domain
To protect validator keys, we're currently using a "proxy key" approach, where the validator key is used to sign a delegation to a new key.  We're also evaluating removing the ability for modules to request signatures from consensus keys altogether in #58 .

An alternative for BLS keys could be to use a different signing domain altogether, which would mean that messages can be signed without risking any slashing. 

Notes:
- if we have a single "commit boost signing domains", then it would be the same across modules (so potentially a signature for module A could be a valid signature for module B)
- we would still need to support proxy keys if we want to support additional schemes, eg ECDSA

ltitanb:
fix #70

Add a commit boost BLS domain used in the signer module. Now signatures on this domain instead of the builder domain

### DIFF
--- a/benches/pbs/src/main.rs
+++ b/benches/pbs/src/main.rs
@@ -4,8 +4,9 @@ use alloy::{primitives::B256, rpc::types::beacon::BlsPublicKey};
 use cb_common::{
     config::RelayConfig,
     pbs::{GetHeaderResponse, RelayClient, RelayEntry},
-    signer::{BlsSecretKey, BlsSigner},
+    signer::BlsSecretKey,
     types::Chain,
+    utils::blst_pubkey_to_alloy,
 };
 use cb_tests::mock_relay::{start_mock_relay_service, MockRelayState};
 use comfy_table::Table;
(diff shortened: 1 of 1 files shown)


crates/common/src/signature.rs
```diff
@@ -76,17 +75,35 @@ pub fn sign_builder_root(
     sign_message(secret_key, &signing_root)
 }
 
+pub fn sign_commit_boost_root(
+    chain: Chain,
+    secret_key: &BlsSecretKey,
+    object_root: [u8; 32],
+) -> BlsSignature {
+    let domain = compute_domain(chain, COMMIT_BOOST_DOMAIN);
+    let signing_root = compute_signing_root(object_root, domain);
```
David-Petrov, line 84: Maybe unite those two lines in a private `sign_for_domain(domain: [u8; 32], secret_key: &BlsSecretKey, object_root: [u8; 32])` which will be called with a specified domain from `sign_builder_root` and `signer_commit_boost_root`?

David-Petrov:
The builder domain is currently also used for the [ECDSA signatures](https://github.com/Commit-Boost/commit-boost-client/blob/d6cfe6655a8f96c4a1b4e0e432755c56edc16616/crates/common/src/signer/schemes/ecdsa.rs#L208) as well, maybe we'd want to update this in the same PR or is this left for a follow-up?

David-Petrov (approved):

PR #112, opened 2024-09-10
feat: optimize dockerfiles

ArchiFleKs:
Signed-off-by: Kevin Lefevre <archi@kiln.fi>


ArchiFleKs:
@mempirate I'll test but can't we also use reqwest with native rust-tls and then use a scratch image with statically compiled binaries? 

I haven't checked the whole codebase but something like this:

```
reqwest = { version = "0.12.5", default-features = false, features = [
  "rustls-tls",
  "json",
] }
```

Should work.

mempirate:
I don't have a lot of experience with scratch images but that could be another option! There are other dependencies like libssl-dev and libssl3 (which include stuff like libcrypto etc) though.

ArchiFleKs:
> I don't have a lot of experience with scratch images but that could be another option! There are other dependencies like libssl-dev and libssl3 (which include stuff like libcrypto etc) though.

Yes, I tried to statically compile with target musl and gnu and changing reqwest feature set to use rusttls-tls, it is almost impossible to get something coherent and it fails due to proc-macros (I somehow managed to build a statically compiled binary on archlinux which had both glibc and musl installed :P but I couldn't reproduced the behavior with docker images and cargo chef) 

I think keeping the binary dynamically linked using debian is the easiest route for now

PR #115, opened 2024-09-11
feat: support environment variables in module config

### LINKED ISSUE #110: Module configuration: support env variables
## Context
The only way to configure modules right now is through the extra config semantics. It would be useful to add options to configure modules with environment variables:
- Add an `env_file` key to the top-level module config
- Add a subtable `[modules.env]` that can specify standalone env values

mempirate:
Closes #110


mempirate:
Ah missed that sorry! Resolved in latest commit

Issue #18, opened 2024-07-16
Add native binary support


ltitanb:
Currently Commit Boost relies heavily on Docker to launch and organize the different modules. This has the advantage of better cross-platform support, sandboxing of modules, and easier integration for node operators which already use docker or k8s. However, this tights us closely to Docker and the Docker API, which not all operators may want to use. 

We need to explore how to support a similar setup without using Docker, by running native binaries instead. 

On linux there are number of internal utilities that can achieve similar sandboxing as docker (and which Docker uses under the hood):
- `seccomp` filters to fine grain syscalls permissions
- PID/NET/MNT namespacing to sandbox processes, network and file system
- control groups (cgroups) to limit memory and disk use
- user permissions 

More research is needed on how to best use these and if there are alternative ways to do it, and how to achieve a similar setup on other platforms eg Windows

orbmis:
Personally I can see that using docker-compose tightly couples the solution to docker, which is a bit of a drawback, but I don't see it as a massive drawback tbh.  I have a background in software development, and I would have always preferred using docker/k8s over other solutions.

Docker is well understood by all devs and works across all platforms, and docker-compose is a neutral declarative format, which can be used to derive k8s manifests, Helm charts, CFN templates, terraform etc. if needed.  I think that supporting docker-compose while also potentially supplying Helm charts could be really useful to a lot of operators.

Using binaries means you will have to make it really easy for devs to compile everything in case they need to, and/or you will also need to compile and ship binaries for well known platforms with every release, which is a bit of a headache for the team that maintains the software.

aimxhaisse:
Related: https://github.com/Commit-Boost/commit-boost-client/issues/81#issuecomment-2304177639

Akagi201:
I use systemd to manage all my services
I found that systemd support sandboxing. These may be useful

https://wiki.archlinux.org/title/User:NetSysFire/systemd_sandboxing
https://wiki.debian.org/ServiceSandboxing

PR #117, opened 2024-09-12
feat: add binary support

ltitanb:
#18 

Allows node operators to run commit boost natively without using Docker:
- tidy up and document the ENV variables that modules need (aka the "contract" between commit boost and modules)
- update docs

Once this is merged, will update the publishing of binaries in the CI on release


David-Petrov (approved):
LGTM 👍

PR #120, opened 2024-09-13
add proxy module check

ltitanb:


crates/signer/src/manager.rs
```diff
@@ -189,6 +189,24 @@ impl SigningManager {
         self.proxy_signers.bls_signers.contains_key(bls_pk)
     }
 
+    pub fn has_proxy_bls_for_module(&self, bls_pk: &BlsPublicKey, module_id: &ModuleId) -> bool {
```
David-Petrov, line 192: Since the notion of a proxy key is coupled with a module ID anyways, I'd delete the old methods:
```rust
    pub fn has_proxy_ecdsa(&self, ecdsa_pk: &EcdsaPublicKey) -> bool {
        self.proxy_signers.ecdsa_signers.contains_key(ecdsa_pk)
    }

    pub fn has_proxy_bls(&self, bls_pk: &BlsPublicKey) -> bool {
        self.proxy_signers.bls_signers.contains_key(bls_pk)
    }
    ```
    
    Quick refactor, they're only used in the tests where a `MODULE_ID` is already present.

David-Petrov (approved):

PR #136, opened 2024-09-26
make get_header generic

fbrv:


fbrv:
@mempirate this change request is still relevant?

mempirate:
It would be, but we would also need the ability to define a generic response. I started working on it but abandoned it due to the amount of changes this approach required to the API. It would still be useful though because we now copy over all of the PBS methods (like `send_timed_get_header`, `send_one_get_header`) just for the ability to receive a different response.

Issue #90, opened 2024-08-21
Support Kurtosis


ltitanb:
We should work to integrate commit boost (starting with the PBS module), with the Kurtosis [ethereum package](https://github.com/ethpandaops/ethereum-package?tab=readme-ov-file#proposer-builder-separation-pbs-emulation). Ideally then we can also run tests with kurtosis in the CI

PR #138, opened 2024-09-27
support chain spec

### LINKED ISSUE #12: Generalize and support different chain `spec` configs
Currently we have hard-coded a few chain specs and IDs, eg [here](https://github.com/Commit-Boost/commit-boost-client/blob/main/crates/common/src/constants.rs) and [here](https://github.com/Commit-Boost/commit-boost-client/blob/main/crates/pbs/src/types/spec.rs).

This should be made generic similar to [ethereum_consensus](https://github.com/ralexstokes/ethereum-consensus/blob/main/ethereum-consensus/src/configs/mod.rs) and ideally be able to load the json returned by [getSpec](https://ethereum.github.io/beacon-APIs/#/Config/getSpec).

ltitanb:
fix #12
also needed for  #90 

add support for different chain specs

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -1,10 +1,10 @@
 # The main configuration file for the Commit-Boost sidecar.
 # Some fields are optional and can be omitted, in which case the default value, if present, will be used.
 
-# Chain spec id. Supported values:
-# - a network id, supported values are: Mainnet, Holesky, Helder
-# - a path to a chain spec file, either in .json (e.g. as returned by the beacon endpoint /eth/v1/config/spec), or .yml format (see examples in tests/data)
-# - a custom object, e.g. chain = { genesis_time_secs = 1695902400, slot_time_secs = 12, genesis_fork_version = "0x01017000" }
+# Chain spec ID. Supported values:
+# A network ID. Supported values: Mainnet, Holesky, Helder.
+# A path to a chain spec file, either in .json format (e.g., as returned by the beacon endpoint /eth/v1/config/spec), or in .yml format (see examples in tests/data).
+# A custom object, e.g., chain = { genesis_time_secs = 1695902400, slot_time_secs = 12, genesis_fork_version = "0x01017000" }.
 chain = "Holesky"
 
 # Configuration for the PBS module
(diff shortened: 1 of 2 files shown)


crates/cli/src/docker_init.rs
```diff
@@ -46,6 +47,17 @@ pub fn handle_docker_init(config_path: String, output_dir: String) -> Result<()>
 
     // config volume to pass to all services
     let config_volume = Volumes::Simple(format!("./{}:{}:ro", config_path, CONFIG_DEFAULT));
+    let chain_spec_volume = chain_spec_path.as_ref().and_then(|p| {
```
fbrv, line 50: maybe:
```rust
let get_file_name_str = |p: &Path| -> Option<&str> {
    p.file_name()?.to_str()
};

// chain_spec_volume using the helper
let chain_spec_volume = chain_spec_path.as_ref().and_then(|p| {
    let file_name = get_file_name_str(p)?;
    Some(Volumes::Simple(format!("{}:/{}:ro", p.display(), file_name)))
});

// chain_spec_env using the helper
let chain_spec_env = chain_spec_path.and_then(|p| {
    let file_name = get_file_name_str(&p)?;
    Some(get_env_val(CHAIN_SPEC_ENV, &format!("/{file_name}")))
});
```

crates/common/src/config/mod.rs
```diff
@@ -45,9 +45,53 @@ impl CommitBoostConfig {
         Ok(config)
     }
 
+    // When loading the config from the environment, it's important that every path
+    // is replaced with the correct value if the config is loaded inside a container
     pub fn from_env_path() -> Result<Self> {
-        let config: Self = load_file_from_env(CONFIG_ENV)?;
```
fbrv, line 49: wonder if we can do something like:
```rust
impl TryFrom<&'static str> for CommitBoostConfig
```
and then
```rust
let config = CommitBoostConfig::try_from(CONFIG_ENV)?;
```
```

fbrv (commented):
in the codebase we are mix-using `*, &, ref(), deref()`, at some point we may want to standardise only one pattern, either *, & or ref(), deref()

fbrv (approved):

PR #149, opened 2024-10-07
chore(typo): `event_publiher` -> `event_publisher`

thedevbirb:
If such PRs are unwanted feel free to close this! Just noticed it while browsing the code


thedevbirb:
> The test is failing because of some false positive serde error, think we dont have every version fully pinned correctly yet. Should be fixed in main if you rebase

Done it! Thanks for the tip. Now the tests pass on local as well.

PR #164, opened 2024-10-23
fix: read body in chunks

ltitanb:
Based on feedback from Sigma Prime auditors:
- we should load the response body by chunks to avoid loading everything in memory at once
- adjust max limits

### DIFF
--- a/crates/pbs/src/utils.rs
+++ b/crates/pbs/src/utils.rs
@@ -12,6 +12,8 @@ pub async fn read_chunked_body_with_max(
     while let Some(chunk) = stream.next().await {
         let chunk = chunk?;
         if response_bytes.len() + chunk.len() > max_size {
+            // avoid spamming logs if the message is too large
+            response_bytes.truncate(1024);
             return Err(PbsError::PayloadTooLarge {
                 max: max_size,
                 raw: String::from_utf8_lossy(&response_bytes).into_owned(),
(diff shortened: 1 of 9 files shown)


crates/pbs/src/utils.rs
```diff
@@ -0,0 +1,25 @@
...
+use reqwest::Response;
+
+pub async fn read_chunked_body_with_max(
+    res: Response,
+    max_size: usize,
+) -> Result<Vec<u8>, PbsError> {
+    let mut stream = res.bytes_stream();
+    let mut response_bytes = Vec::new();
+
+    while let Some(chunk) = stream.next().await {
```
fbrv, line 12: wonder if we can do something like so we avoid variables
```rust
pub async fn read_chunked_body_with_max(
    res: Response,
    max_size: usize,
) -> Result<Vec<u8>, PbsError> {
    let response_bytes = res
        .bytes_stream()
        .map_err(|e| PbsError::Reqwest(e))
        .try_fold(Vec::new(), |mut acc, chunk| async move {
            if acc.len() + chunk.len() > max_size {
                Err(PbsError::PayloadTooLarge {
                    max: max_size,
                    raw: String::from_utf8_lossy(&acc).into_owned(),
                })
            } else {
                acc.extend_from_slice(&chunk);
                Ok(acc)
            }
        })
        .await?;

    Ok(response_bytes)
}

```

fbrv (approved):

Issue #71, opened 2024-08-08
Persist proxy keys


ltitanb:
Modules can request proxy keys from the signing manager. How should the proxy private keys be stored and to what extent we can leverage remote signers for this (considering also that the key might not necessarily be BLS)

PR #166, opened 2024-10-30
feat(signer): store proxy keys and delegations

ltitanb:
First implementation of #71 . Proxy keys and delegations are stored in plain text for testing purposes


David-Petrov (approved):
LGTM 👍

Issue #15, opened 2024-07-16
Add `BuilderApi` multiplexer


ltitanb:
We should support use cases where users need to send some requests in the builder API to different sidecars based on the validator pubkey, which allows users to run a single beacon client (which generally support a single BuilderAPI endpoint). 

In practice this should be a mapping of `Array<BlsPublicKey>` -> `PbsModuleConfig` in the main config file, so that users can customize eg. relays and different images to run.

aimxhaisse:
Bunch of ideas for more advanced filtering:

- by lido operator id (i.e: commit-boost would fetch keys from the Lido smart-contract) ; this is useful because Lido has as a whitelisted set of relayers to use,
- by SSV operator id (i.e: commit-boost would fetch keys from the SSV smart-contract),
- by withdrawal credentials (i.e: to have a specific customer target specific relays).

Tricky part of this is: how to keep the mapping updated, for instance, if a new key is added to a setup ; this implies to have a way to automatically reload the mapping in real-time (or periodically restart commit-boost and accept a lag).

aimxhaisse:
Example of such a "hypothetical" config:

```yml
      strategies:
        - name: default
          relays: ['holesky-titan', 'holesky-bloxroute', 'holesky-aestus', 'holesky-eden', 'holesky-ultrasound']
        - name: lido-holesky
          relays: ['holesky-flashbots']
          filters:
            lido_operators: [8]
        - name: lido-ssv
          relays: ['holesky-titan', 'holesky-flashbots', 'holesky-eden']
          filters:
            ssv_operators: [199]
        - name: primev
          relays: ['holesky-titan']
          delaying: False
          filters:
            public_keys:
              - 88f51c0193d5b3f0ff408ea82e57ba9d1c19206b6afa7a55fde4580f326cc055fb407fdf37bdb22bf6ddf533ec218e7d
              - 88f51e63fb82120c9a405b4cb03e45d922a91884fcaec3ef2426eecf25c70695dca8806731a082b746fe4673a28b2155
              - 88f56fa47d863ecc1050eaa9eae86bbc1987490d8b7a66cf9bf69727aea603b3ca90bdbb2d79de8411cea280838f6800
              - 88f598d3ff3a1c9659f795d851c79dafd6c3bac12f55f87221da3127059b4223bb0398ee599802351ab95134baa7b318
[...]

PR #172, opened 2024-11-19
feat: add pbs mux

ltitanb:
add a PBS mux to override the config when calling get header , currently only supporting a hardcoded set of validator pubkeys

#15

### DIFF
--- a/crates/common/src/config/mux.rs
+++ b/crates/common/src/config/mux.rs
@@ -41,7 +41,7 @@ impl PbsMuxes {
 
         let mut configs = HashMap::new();
         // fill the configs using the default pbs config and relay entries
-        for mux in self.muxes.into_iter() {
+        for mux in self.muxes {
             ensure!(!mux.relays.is_empty(), "mux config must have at least one relay");
             ensure!(
                 !mux.validator_pubkeys.is_empty(),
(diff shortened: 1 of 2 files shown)


crates/common/src/config/mux.rs
```diff
@@ -0,0 +1,138 @@
...
+            for pubkey in mux.validator_pubkeys.iter() {
+                if !unique_pubkeys.insert(pubkey) {
+                    bail!("duplicate validator pubkey in muxes: {pubkey}");
+                }
+            }
+        }
+
+        let mut configs = HashMap::new();
+        // fill the configs using the default pbs config and relay entries
+        for mux in self.muxes.into_iter() {
```
fbrv, line 44: it's not the same as 
```rust
for mux in self.muxes
```
?

crates/common/src/config/pbs.rs
```diff
@@ -110,7 +120,12 @@ fn default_pbs() -> String {
 /// Loads the default pbs config, i.e. with no signer client or custom data
 pub fn load_pbs_config() -> Result<PbsModuleConfig> {
     let config = CommitBoostConfig::from_env_path()?;
+    config.validate()?;
 
+    let muxes = match config.muxes {
```
fbrv, line 125: a bit more idiomatic
```rust
let muxes = config.muxes.map(|muxes| {
    muxes.validate_and_fill(&config.pbs.pbs_config, &config.relays)
}).transpose()?;
```

fbrv (approved):

PR #187, opened 2024-11-22
CBST-02: add retries

ltitanb:


crates/pbs/src/mev_boost/register_validator.rs
```diff
@@ -63,15 +64,61 @@ pub async fn register_validator<S: BuilderApiState>(
...
     relay: RelayClient,
     headers: HeaderMap,
     timeout_ms: u64,
 ) -> Result<(), PbsError> {
     let url = relay.register_validator_url()?;
+    let mut remaining_timeout_ms = timeout_ms;
+    let mut retry = 0;
+    let mut backoff = Duration::from_millis(250);
+
+    loop {
```
fbrv, line 80: can we use the reqwest middleware?

https://docs.rs/reqwest-retry/latest/reqwest_retry/

PR #188, opened 2024-11-22
CBST-03: validate more get header

ltitanb:


crates/pbs/Cargo.toml
```diff
@@ -37,4 +38,4 @@ thiserror.workspace = true
 eyre.workspace = true
 url.workspace = true
 uuid.workspace = true
-lazy_static.workspace = true
\ No newline at end of file
+lazy_static.workspace = true
```
fbrv, line 41: for future PR, let's consider to deprecate lazy_static and use once_cell if the ux is easy enough

fbrv (approved):

PR #207, opened 2024-12-11
feat: add the ability to skip the relayer public key check

aimxhaisse:


aimxhaisse:
> looks good! can you please update the description in the example config, that `skip_sigverify` also skips the public key check?

Done

PR #212, opened 2024-12-16
feat(pbs): stateless pbs module

### LINKED ISSUE #61: Consider a stateless design on the PBS module
When operating at scale it is convenient to have a stateless approach where you can scale the number of commit-boost instances blindly:

This allows to setup an ingress used by all beacons that spreads requests to all commit-boost instances in round-robin or more evolved forms. This has the following advantages:

- Scales well the registration calls: if you have a large infra (for example operating 100 000 validators on Holesky), you don't want a single instance of commit-boost to handle all registrations because it will be a bottleneck: relayers often return rate-limited 4XX upon large load, so you spreading from multiple IPs (instances) is needed.

- Operationally speaking it is easy to manage: you can upgrade one by one your instances with 0 downtime, canary some to qualify behavior, etc.

It has the following drawbacks on the design-side: PBS would have to be stateless, with the implication that a blinded block call for instance can land on a different instances that returned the header. This means the commit-boost instance can't know from which relayer it got the block: this is not an issue because to improve global propagation it is convenient to send the block anyway to all relayers (more than one can know about the block).

This kind of code becomes not possible with this approach: https://github.com/Commit-Boost/commit-boost-client/blob/main/crates/pbs/src/routes/submit_block.rs#L66

To me the stateless approach works very well at scale with PBS, I don't see how we could operate properly without this. However I'm unsure how it'd work for other modules.

ltitanb:
fix #61

Transition the default PBS module to a fully stateless setup. In practice this means we lose:
- the `X-MEVBoost-SlotID` header that linked getHeader/getPayload calls
- logs on missed payload of all relays that delivered the header but not the paylod (this is still accessible from previous logs or via the relay data API)

In exchange we get a cleaner and more scalable setup, see also the discussions in #61 and [CC2](https://github.com/Commit-Boost/pm/blob/main/Community_Calls/Call2.md)

After this the `PbsState` is effectively just a wrapper for the config plus any state stored by modules that is left for now to avoid breaking anything downstream. We should evaluate separately if this should warrant a redesign of PBS custom modules

### DIFF
--- a/crates/common/src/config/mux.rs
+++ b/crates/common/src/config/mux.rs
@@ -216,10 +216,7 @@ async fn fetch_lido_registry_keys(
     chain: Chain,
     node_operator_id: U256,
 ) -> eyre::Result<Vec<BlsPublicKey>> {
-    debug!(
-        "loading operator keys from Lido registry: chain={:?}, node_operator_id={}",
-        chain, node_operator_id
-    );
+    debug!(?chain, %node_operator_id, "loading operator keys from Lido registry");
 
     let provider = ProviderBuilder::new().on_http(rpc_url);
     let registry_address = lido_registry_address(chain)?;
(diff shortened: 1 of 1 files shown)


ManuelBilbao (approved):
LGTM!

PR #227, opened 2025-01-09
fix: use all relays for some calls

### LINKED ISSUE #220: use all relays union for registrations
After #205 , relays can be defined in a mux config without being in the `[[relays]]` section. 
Currently we use the mux only for get header, and `[[relays]]` for other calls, but the correct behaviour is to take the union of all relays (in mux or not) for those calls

ltitanb:
fix #220

Use all relays (default + mux) in all calls except get_header


ManuelBilbao (approved):
LGTM!

Issue #218, opened 2024-12-24
[FEAT] Add num_registrations to log and metrics


blevkivskyi-everstake:
I have two muxes with different relays, and it would be good to have num_registrations for each of them in the log, for example.

From this:
```bash
{"timestamp":"2024-12-23T14:33:36.328688Z","level":"DEBUG","fields":{"message":"registration successful","code":"200","latency":"296.08105ms"},"target":"cb_pbs::mev_boost::register_validator","spans":[{"req_id":"353bdfc2-5a0f-46a6-8027-6b03d632a93d","name":"register_validators"},{"relay_id":"boost-relay-holesky.flashbots.net","retry":0,"name":"handler"}]}

{"timestamp":"2024-12-23T14:33:36.438698Z","level":"DEBUG","fields":{"message":"registration successful","code":"200","latency":"406.05268ms"},"target":"cb_pbs::mev_boost::register_validator","spans":[{"req_id":"353bdfc2-5a0f-46a6-8027-6b03d632a93d","name":"register_validators"},{"relay_id":"bloxroute.holesky.blxrbdn.com","retry":0,"name":"handler"}]}
```
To this:
```bash
{"timestamp":"2024-12-23T14:33:36.328688Z","level":"DEBUG","fields":{"message":"registration successful","code":"200","latency":"296.08105ms", "num_registrations":3},"target":"cb_pbs::mev_boost::register_validator","spans":[{"req_id":"353bdfc2-5a0f-46a6-8027-6b03d632a93d","name":"register_validators"},{"relay_id":"boost-relay-holesky.flashbots.net","retry":0,"name":"handler"}]}

{"timestamp":"2024-12-23T14:33:36.438698Z","level":"DEBUG","fields":{"message":"registration successful","code":"200","latency":"406.05268ms", "num_registrations":3},"target":"cb_pbs::mev_boost::register_validator","spans":[{"req_id":"353bdfc2-5a0f-46a6-8027-6b03d632a93d","name":"register_validators"},{"relay_id":"bloxroute.holesky.blxrbdn.com","retry":0,"name":"handler"}]}
```

Issue #226, opened 2025-01-09
[FEAT] Add new config parameter to set batch-size for registration validators on relays


blevkivskyi-everstake:
To improve performance when dealing with a high number of keys, we propose adding a new configuration parameter to the mux config like [lighthouse validator](https://lighthouse-book.sigmaprime.io/help_vc.html?highlight=batc#validator-client:~:text=%2D%2Dvalidator%2Dregistration%2Dbatch%2Dsize).

A error when sending a large number of keys to a relay will be fixed::

```bash
HTTP error response: 413 / Failed to buffer the request body: length limit exceeded
```

Issue #225, opened 2025-01-09
[FEAT] Add more parameter to one log line


blevkivskyi-everstake:
It would be great to have in **one** log line more parameters, that could be processed by Grafana/LOKI

Commit-boost:
```
2025-01-01T04:04:00.405411Z  INFO get_header{req_id=2869e01c-a39d-4793-805c-29067db47846 slot=3316820}: received header value_eth="0.002629109072852370" block_hash=0x8a4873e7d485d56fc28bd4d15ed21f259b243b6449a4fab71561693fe208d288

2025-01-01T04:04:00.394390Z DEBUG get_header{req_id=2869e01c-a39d-4793-805c-29067db47846 slot=3316820}:handler{relay_id="boost-relay-holesky.flashbots.net"}: received new header latency=262.760568ms value_eth="0.002629109072852370" block_hash=0x8a4873e7d485d56fc28bd4d15ed21f259b243b6449a4fab71561693fe208d288

2025-01-01T04:04:00.131335Z DEBUG get_header{req_id=2869e01c-a39d-4793-805c-29067db47846 slot=3316820}: using mux config mux_id="test_mux_2" relays=1 pubkey=0xb28dbbcec926781dcc028c0601ffbb0df6316ba41fac800ffd650e9b8b00414fc34a0c77ba78299883aab9254ca1554a

2025-01-01T04:04:00.131283Z  INFO get_header{req_id=2869e01c-a39d-4793-805c-29067db47846 slot=3316820}: ua="Lighthouse/v6.0.1-0d90135" parent_hash=0x6f1c01d150fed6217b5a4e5c026e5615d957177d4e3e22174c94740ffef3b277 validator_pubkey=0xb28dbbcec926781dcc028c0601ffbb0df6316ba41fac800ffd650e9b8b00414fc34a0c77ba78299883aab9254ca1554a ms_into_slot=131
```

Mev-boost:
```
time="2025-**-**T**:**:**.***Z" level=info msg="best bid" blockHash=0x*** blockNumber=*** method=getHeader parentHash=0x*** pubkey=0x*** relays="http://0x***" slot=*** slotUID=***-***-***-***-*** txRoot=0x*** ua=teku/v24.12.1 value=100000000000 version=v1.8.1
```

This will be useful for Grafana alerting:

 ![image](https://github.com/user-attachments/assets/24802ac2-d3e9-4b02-94c4-4d016a69ace3)

[...]

PR #228, opened 2025-01-09
feat(mux): add successful num_registrations to logs

ManuelBilbao:
For each validator, log the number of relays where it was successfully registered, if `wait_all_registrations` config is `true`.

#218 

> [!WARNING]
> This changes the Eth Builder API in order to return that number. If this is not ideal, we can otherwise log the number of relays where the validator **attempted** to register

### DIFF
--- a/crates/pbs/src/api.rs
+++ b/crates/pbs/src/api.rs
@@ -45,7 +45,7 @@ pub trait BuilderApi<S: BuilderApiState>: 'static {
         registrations: Vec<ValidatorRegistration>,
         req_headers: HeaderMap,
         state: PbsState<S>,
-    ) -> eyre::Result<()> {
+    ) -> eyre::Result<usize> {
         mev_boost::register_validator(registrations, req_headers, state).await
     }
 }
--- a/crates/pbs/src/mev_boost/register_validator.rs
+++ b/crates/pbs/src/mev_boost/register_validator.rs
@@ -25,7 +25,7 @@ pub async fn register_validator<S: BuilderApiState>(
     registrations: Vec<ValidatorRegistration>,
     req_headers: HeaderMap,
     state: PbsState<S>,
-) -> eyre::Result<()> {
+) -> eyre::Result<usize> {
     // prepare headers
     let mut send_headers = HeaderMap::new();
     send_headers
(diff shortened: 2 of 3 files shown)


ManuelBilbao:
@ltitanb not sure what you mean in the last item.
My understanding is that the module receives a request and resends it to many relays. So the number of successful responses will be the number of relays that successfully register the request's validators.

Also, I changed the logs so now they are equal in both cases. If `wait_all_registrations` is disabled, it will throw another async task to wait all requests and finally log the result. Let me know if that was the expected behaviour

ank-everstake:
Please see log example here:
https://github.com/Commit-Boost/commit-boost-client/issues/225#:~:text=Proposed%20json%20format
```
num_registrations will mean amount of keys (1000/1000/3000)
num_responses will mean 10/5/3
```

We would like to see number of validator keys registered in **each** relay.
Also we want to ask for **batching** as relays fail to register if number of keys in request is high.

Linked feature-request:
https://github.com/Commit-Boost/commit-boost-client/issues/226

ManuelBilbao:
Currently we use the mux only for `get_header` requests, so the `register_validator` request will send all validators to all relays.
Following your example from #226 , I setup a config with:
```
1000 keys
5 relays:
    700 allowed keys to use relay-2 and relay-3
    300 allowed keys to use relay-4 and relay-5
    A default relay-1
```
And sent a `register_validator` request with all 1000 keys. The logs results were:

stdout:
```
2025-01-17T15:33:33.318870Z  INFO register_validators{req_id=fb286e5c-3c59-4561-b395-a8bebaddffc2}: all relay registrations finished num_relays=5 num_registrations=1000 total_time=12.471209ms
```

JSON
```json
{"timestamp":"2025-01-17T15:33:33.318885Z","level":"INFO","fields":{"message":"all relay registrations finished","num_relays":5,"num_registrations":1000,"total_time":"12.471209ms"},"target":"cb_pbs::mev_boost::register_validator","spans":[{"req_id":"fb286e5c-3c59-4561-b395-a8bebaddffc2","name":"register_validators"}]}
```

Let me know if this is ok or you need something different

ank-everstake:
Could you share `DEBUG` level logs for this setup to double-check?

PR #234, opened 2025-01-14
feat: add info metrics

### LINKED ISSUE #233: [FEAT] PBS module version exposed as metric
It would be great to have the version of the PBS module exposed as a metric (other future commit-boost modules too).

**Why?**
We run multiple instances of `commit-boost-pbs`. When rolling out updates across them, we can't easily tell which ones still need to be updated or double check that the update was indeed successful.

(It would also be good to show this information on the public PBS Grafana dashboard)

Some examples of how other teams do this:

`beacon_teku_version{[...labels...], version="24.1.1"}`

`lodestar_version{commit="5d93a629c0fd1fdd32160cbf7717e7a6b22f7f2d", [...], version="v1.15.0/5d93a62"}}`

`nethermind_version{Commit="4f89be99ef7d7cef7f63f4742e4222279c161247", Version="1.26.0-exp.2+4f89be99"}`

`vero_info{commit="d2a45ba951c2a27e6816e7d9e9f67f2d1b217026", version="0.8.3"}`

This keeps it really simple. It seems the convention/best practice is to set this metric's value to 1.

ltitanb:
fix #233 

add info metric and panel

image

### DIFF
--- a/crates/common/Cargo.toml
+++ b/crates/common/Cargo.toml
@@ -4,7 +4,6 @@ version.workspace = true
 edition.workspace = true
 rust-version.workspace = true
 publish = false
-build = "build.rs"
 
 [dependencies]
 # ethereum
--- a/examples/da_commit/src/main.rs
+++ b/examples/da_commit/src/main.rs
@@ -105,11 +105,12 @@ async fn main() -> Result<()> {
     // Remember to register all your metrics before starting the process
     MY_CUSTOM_REGISTRY.register(Box::new(SIG_RECEIVED_COUNTER.clone()))?;
     // Spin up a server that exposes the /metrics endpoint to Prometheus
-    MetricsProvider::load_and_run(Chain::Mainnet, MY_CUSTOM_REGISTRY.clone())?;
 
     match load_commit_module_config::<ExtraConfig>() {
         Ok(config) => {
             let _guard = initialize_tracing_log(&config.id)?;
+
+            MetricsProvider::load_and_run(config.chain, MY_CUSTOM_REGISTRY.clone())?;
 
             info!(
                 module_id = %config.id,


crates/common/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version.workspace = true
 edition.workspace = true
 rust-version.workspace = true
 publish = false
+build = "build.rs"
```
ManuelBilbao, line 7: nit: this isn't needed as it's the default value

crates/common/build.rs
```diff
@@ -0,0 +1,7 @@
+use std::process::Command;
+
+fn main() {
+    let output = Command::new("git").args(["rev-parse", "HEAD"]).output().unwrap();
+    let git_hash = String::from_utf8(output.stdout).unwrap();
+    println!("cargo:rustc-env=GIT_HASH={}", git_hash);
```
ManuelBilbao, line 6: Nit
```suggestion
    println!("cargo:rustc-env=GIT_HASH={git_hash}");
```

crates/signer/proto
ManuelBilbao, line 1: I think this shouldn't be in this PR

crates/common/src/types.rs
```diff
@@ -28,6 +28,15 @@ pub enum Chain {
 
 pub type ForkVersion = [u8; 4];
 
+impl std::fmt::Display for Chain {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Mainnet | Self::Holesky | Self::Sepolia | Self::Helder => write!(f, "{:?}", self),
```
ManuelBilbao, line 34: Nit
```suggestion
            Self::Mainnet | Self::Holesky | Self::Sepolia | Self::Helder => write!(f, "{self:?}"),
```

examples/da_commit/src/main.rs
```diff
@@ -105,7 +105,7 @@ async fn main() -> Result<()> {
     // Remember to register all your metrics before starting the process
     MY_CUSTOM_REGISTRY.register(Box::new(SIG_RECEIVED_COUNTER.clone()))?;
     // Spin up a server that exposes the /metrics endpoint to Prometheus
-    MetricsProvider::load_and_run(MY_CUSTOM_REGISTRY.clone())?;
+    MetricsProvider::load_and_run(Chain::Mainnet, MY_CUSTOM_REGISTRY.clone())?;
```
ManuelBilbao, line 108: Should this be inside the `Ok(config)` arm so that the chain is obtained from the config?

ManuelBilbao (approved):
Looks good! Just left some nit comments

ManuelBilbao (approved):

PR #252, opened 2025-02-05
feat(pbs): add builder SSZ flow

eserilev:
Spec: https://github.com/ethereum/builder-specs/pull/104

Right now both mev-boost and commit-boost don't support the new SSZ builder flow. Theres already some effort being done to get mev-boost updated, so thought I'd go ahead and add SSZ to commit-boost. Lighthouse unstable already supports the new SSZ flow, i'll try testing locally and report back with the results of those tests.

It'd also be really really nice if we could get electra support in commit-boost. Not sure if any progress has been made on that front, if not I can try tackling that next. Ideally we'd love multiple builder implementations to test on electra testnets before the hardfork


crates/pbs/src/routes/submit_block.rs
```diff
@@ -39,9 +42,31 @@ pub async fn handle_submit_block<S: BuilderApiState, A: BuilderApi<S>>(
             trace!(?res);
             state.publish_event(BuilderEvent::SubmitBlockResponse(Box::new(res.clone())));
             info!("received unblinded block");
-
             BEACON_NODE_STATUS.with_label_values(&["200", SUBMIT_BLINDED_BLOCK_ENDPOINT_TAG]).inc();
-            Ok((StatusCode::OK, Json(res).into_response()))
+
+            let response = match content_type_header {
```
jtraglia, line 47: Hey @eserilev, this response type should be based on the Accept header value, like we do in `handle_get_header`.

For reference, see this line in the spec: https://github.com/ethereum/builder-specs/blob/1bbe1e299aa174aa6b7ec7a84c59c678b4e1b063/apis/builder/blinded_blocks.yaml#L83

> Use Accept header to choose this response type

crates/common/src/utils.rs
```diff
@@ -273,6 +281,158 @@ pub fn get_user_agent_with_version(req_headers: &HeaderMap) -> eyre::Result<Head
...
+        req_headers.get(CONTENT_TYPE).and_then(|value| value.to_str().ok()).unwrap_or("application/json"),
+    )
+    .unwrap_or(ContentType::Json)
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum ContentType {
+    Json,
+    Ssz
+}
```
jtraglia, line 304: Just a nit, but I would rename this enum to `MediaType` which is most accurate term.

crates/pbs/src/routes/get_header.rs
```diff
@@ -33,17 +34,37 @@ pub async fn handle_get_header<S: BuilderApiState, A: BuilderApi<S>>(
...
                 info!(value_eth = format_ether(max_bid.value()), block_hash =% max_bid.block_hash(), "received header");
-
                 BEACON_NODE_STATUS.with_label_values(&["200", GET_HEADER_ENDPOINT_TAG]).inc();
-                Ok((StatusCode::OK, axum::Json(max_bid)).into_response())
+                let response = match accept_header {
+                    Accept::Ssz => {
+                        let mut res = {
+                            info!("sending response as JSON");
+                            (StatusCode::OK, max_bid.data.as_ssz_bytes()).into_response()
+                        };
```
jtraglia, line 51: I'm a little confused by this log message. It's wrong, right?

crates/common/src/utils.rs
jtraglia: Well actually, I understand what you're doing now. You're splitting this into `ContentType` and `Accept`. I would probably combine these into a `MediaType` enum and try to not have an `Any` value.

crates/common/src/utils.rs
```diff
@@ -273,6 +281,158 @@ pub fn get_user_agent_with_version(req_headers: &HeaderMap) -> eyre::Result<Head
...
+                let payload: T = serde_json::from_slice(&bytes)
+                    .map_err(|_| StatusCode::BAD_REQUEST.into_response())?;
+                return Ok(Self(payload));
+            }
+
+            if content_type.starts_with(&ContentType::Ssz.to_string()) {
+                let payload = T::from_ssz_bytes(&bytes)
+                    .map_err(|_| StatusCode::BAD_REQUEST.into_response())?;
+                return Ok(Self(payload));
+            }
```
jtraglia, line 428: I must be missing something. Where in this PR are we reading the `Eth-Consensus-Version` header value? I do not see a `headers.get(CONSENSUS_VERSION_HEADER)` call anywhere.

eserilev: Commit boost only works for Deneb at the moment so the consensus version header isn't relevant. They are working on supporting Electra in a separate PR. That PR should be adding logic to read the consensus version header

crates/pbs/src/routes/submit_block.rs
eserilev: thanks for catching that, this should be fixed now

eserilev:
I've added some integration tests for the new ssz flow. I haven't had a chance to test this in kurtosis yet. I was having some difficulty running commit-boost in kurtosis (some missing config error). I'll try again this week and if I have issues I'll reach out on TG

eserilev:
im working on resolving merge conflicts

PR #255, opened 2025-02-10
Add support for fetching keys for mux from HTTP source

kamsz:
Hey,

This PR adds support for a simple HTTP endpoint loader for mux. It expects the same format of data as text file (array of pubkeys).


kamsz:
> Thanks! Could you please add an example format for the pubkeys returned?

Thanks, added a comment with example response to configuration file.

PR #300, opened 2025-04-29
CBST2-11: Improve path validation on local Mux keys loader

ManuelBilbao:
Check if the passed file exists and has `.json` extension


crates/common/src/config/mux.rs
```diff
@@ -120,18 +120,28 @@ pub struct MuxConfig {
 impl MuxConfig {
     /// Returns the env, actual path, and internal path to use for the file
     /// loader
```
jclapis, line 122:
```suggestion
    /// loader. In File mode, validates the mux file prior to returning.   
```

jclapis (approved):

PR #340, opened 2025-07-21
Add Nimbus loader support

jclapis:
This adds support for Nimbus keys to the signer service's key loader. Nimbus's format is almost identical to Lighthouse's, it just has a different keystore filename.


jclapis:
> Looks good. We should add the format in `docs/docs/get_started/configuration.md` too (see https://commit-boost.github.io/commit-boost-client/get_started/configuration/#local-signer)

Done, thanks for the heads up!

ManuelBilbao (approved):

PR #353, opened 2025-08-06
Change chain IDs to 256-bit numbers

jclapis:
Currently a size limit on chain IDs isn't well-defined (see https://ethereum-magicians.org/t/eip-2294-explicit-bound-to-chain-id). Some tools use an unsigned 64-bit integer (including `alloy` which is one of our dependencies), some use an unsigned 52-bit integer (since that's the highest value that fits without error into floats, which JS uses), but the largest value canonically is an unsigned 256-bit integer, used by Geth and friends in the EVM implementation itself.

One of the reasons the 256-bit version is popular is because it lets people use `keccak256("some_relevant_string")` as their Chain ID to avoid collisions. Until there's an official spec designation for a size limit, we should stick to 256-bit ones internally to support use-cases like that.

Alternatively, we should make it explicit in the documentation that Commit-Boost doesn't support chain IDs larger than the `u64` limit; people will not be able to use e.g. the Signer service for custom chains that go out of those bounds.


crates/common/src/utils.rs
```diff
@@ -443,6 +443,22 @@ pub fn get_user_agent_with_version(req_headers: &HeaderMap) -> eyre::Result<Head
...
+impl FromU64 for B256 {
+    /// Converts a u64 into a B256 by left-padding with zeros
+    fn from_u64(value: u64) -> B256 {
+        let mut buffer = [0u8; 32];
+        let bytes = value.to_be_bytes();
+        buffer[24..].copy_from_slice(&bytes);
+        B256::from(buffer)
+    }
+}
+
```
ManuelBilbao, line 461: Do we really need this? A conversion can be done by simply `B256::from(U256::from(u64_value))`

jclapis: I'm maintaining 2 branches in parallel - one with this and one without this, and that snuck in from the branch without it. I'll remove it from this one, thanks for the suggestion.

ManuelBilbao (approved):

PR #356, opened 2025-08-12
Add payload hash to signer JWT claims

jclapis:
## :warning: Do not merge unntil #354 and #353 are in!

This is part 2 of the update on CBST2-01, following #354 and #353. This solves one of the issues found within the audit by making all of the routes with a request body (e.g., the `POST` routes) encode the Keccak256 hash of the payload body into the JWT claims for the request's auth header. Doing so means JWTs can't be intercepted and reused for unrelated requests, such as for signing different things other than what the original request was for. This affects all routes, including the new `/revoke_jwt` and `/reload` ones.


crates/signer/src/service.rs
```diff
@@ -224,42 +238,83 @@ fn check_jwt_rate_limit(state: &SigningState, client_ip: &IpAddr) -> Result<(),
...
+            SignerModuleError::Unauthorized
+        })?;
+    } else {
+        validate_jwt(jwt, &jwt_config.jwt_secret, Some(body)).map_err(|e| {
+            error!("Unauthorized request. Invalid JWT: {e}");
+            SignerModuleError::Unauthorized
+        })?;
+
+        // Make sure the request contains a hash of the payload in its claims
+        if !body.is_empty() {
```
ManuelBilbao, line 271: Won't this always be true?

api/signer-api.yml
```diff
@@ -593,6 +633,185 @@ paths:
               schema:
                 type: string
                 example: "OK"
+  /reload:
```
ManuelBilbao, line 636: We discussed some time ago if we should include non Signer API endpoints here or not. We can discuss it again though, if you have another point of view

Ref: https://github.com/Commit-Boost/commit-boost-client/pull/231#discussion_r1924129461

crates/common/src/commit/client.rs
```diff
@@ -115,10 +128,11 @@ impl SignerClient {
         Q: Serialize,
         T: for<'de> Deserialize<'de>,
     {
-        self.refresh_jwt()?;
+        let auth_header = self.create_jwt_header_for_payload(request)?;
 
         let url = self.url.join(route)?;
-        let res = self.client.post(url).json(&request).send().await?;
+        let res =
+            self.client.post(url).json(&request).header(AUTHORIZATION, auth_header).send().await?;
```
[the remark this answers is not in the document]
jclapis: Nice, cleaned up in f25f6fe.

PR #380, opened 2025-09-22
Add a task to clean up expired JWT failure rate limits 

jclapis:
This adds a small task that runs in parallel with the Signer server and periodically cleans up the JWT auth failure map, removing any rate limiting entries that have expired. Previously the only way they got removed was if the timed out client tried to request again after the timeout period, so if lots of clients never re-requested then they'd just fill up the map for no reason.

Note that the attempt count isn't timed, so if a client fails 2/3 attempts, then waits 10 minutes, then fails a 3rd attempt, they'll trigger the lockout.

### DIFF
--- a/crates/common/src/config/signer.rs
+++ b/crates/common/src/config/signer.rs
@@ -88,7 +88,8 @@ pub struct SignerConfig {
     pub jwt_auth_fail_limit: u32,
 
     /// Duration in seconds to rate limit an endpoint after the JWT auth failure
-    /// limit has been reached
+    /// limit has been reached. This also defines the interval at which failed
+    /// attempts are regularly checked and expired ones are cleaned up.
     #[serde(default = "default_u32::<SIGNER_JWT_AUTH_FAIL_TIMEOUT_SECONDS_DEFAULT>")]
     pub jwt_auth_fail_timeout_seconds: u32,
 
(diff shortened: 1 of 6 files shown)


crates/common/src/config/signer.rs
```diff
@@ -88,7 +88,8 @@ pub struct SignerConfig {
     pub jwt_auth_fail_limit: u32,
 
     /// Duration in seconds to rate limit an endpoint after the JWT auth failure
-    /// limit has been reached
+    /// limit has been reached. This also defines the interval at which failed
+    /// attempts are regularly checked and expired ones are cleaned up.
```
ManuelBilbao, line 92: Should we also add this note to `config.example.toml`?

crates/signer/src/service.rs
```diff
@@ -143,6 +143,23 @@ impl SigningService {
             .route_layer(middleware::from_fn(log_request))
             .route(STATUS_PATH, get(handle_status));
 
+        // Run the JWT cleaning task
+        //let state = state.clone();
```
ManuelBilbao, line 147: I think this is missed here

tests/Cargo.toml
```diff
@@ -21,3 +21,6 @@ tracing.workspace = true
 tracing-subscriber.workspace = true
 tree_hash.workspace = true
 url.workspace = true
+
+[dev-dependencies]
+tracing-test.workspace = true
```
ManuelBilbao, line 26: Nit.
```suggestion
tracing-test.workspace = true

```

PR #388, opened 2025-10-01
refactor(signer): Use config-based header to extract IP from

ManuelBilbao:
Add a config param to set the trusted HTTP header to get the user's IP from

### DIFF
--- a/config.example.toml
+++ b/config.example.toml
@@ -173,9 +173,9 @@ jwt_auth_fail_timeout_seconds = 300
 # HTTP header to use to determine the real client IP, if the Signer is behind a proxy (e.g. nginx)
 # OPTIONAL. If missing, the client IP will be taken directly from the TCP connection.
 # [signer.reverse_proxy]
-# Unique: HTTP header name to use to determine the real client IP. Expected to appear only once in the request.
+# Unique: HTTP header name to use to determine the real client IP. Expected to appear only once in the request. Requests with multiple values of this header will be rejected.
 # unique = "X-Real-IP"
-# Rightmost: HTTP header name to use to determine the real client IP from a comma-separated list of IPs. Rightmost IP is the client IP.
+# Rightmost: HTTP header name to use to determine the real client IP from a comma-separated list of IPs. Rightmost IP is the client IP. If the header appears multiple times, the last value will be used.
 # rightmost = "X-Forwarded-For"
 
 # [signer.tls_mode]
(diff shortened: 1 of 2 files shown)


docs/docs/get_started/configuration.md
```diff
@@ -399,11 +399,21 @@ jwt_auth_fail_limit = 3 # The amount of failed requests allowed
 jwt_auth_fail_timeout_seconds = 300 # The time window in seconds
 ```
 
-The rate limit is applied to the IP address of the client making the request. By default, the IP is extracted directly from the TCP connection. If you're running the Signer service behind a reverse proxy (e.g. Nginx), you can configure it to extract the IP from a custom HTTP header instead:
+The rate limit is applied to the IP address of the client making the request. By default, the IP is extracted directly from the TCP connection. If you're running the Signer service behind a reverse proxy (e.g. Nginx), you can configure it to extract the IP from a custom HTTP header instead. There're two options:
+
+- `unique`: The name of the HTTP header that contains the IP. This header is expected to appear only once in the request. This is common when using `X-Real-IP`, `True-Client-IP`, etc.
```
jclapis:
```suggestion
- `unique`: The name of the HTTP header that contains the IP. This header is expected to appear only once in the request. This is common when using `X-Real-IP`, `True-Client-IP`, etc. If a request is received that has multiple values for this header, it will be considered invalid and rejected.
```

config.example.toml
```diff
@@ -169,9 +169,14 @@ jwt_auth_fail_limit = 3
...
 # OPTIONAL, DEFAULT: 300
 jwt_auth_fail_timeout_seconds = 300
+
 # HTTP header to use to determine the real client IP, if the Signer is behind a proxy (e.g. nginx)
 # OPTIONAL. If missing, the client IP will be taken directly from the TCP connection.
-# trusted_ip_header = "X-Real-IP"
+# [signer.reverse_proxy]
+# Unique: HTTP header name to use to determine the real client IP. Expected to appear only once in the request.
+# unique = "X-Real-IP"
+# Rightmost: HTTP header name to use to determine the real client IP from a comma-separated list of IPs. Rightmost IP is the client IP.
```
jclapis:
```suggestion
# Rightmost: HTTP header name to use to determine the real client IP from a comma-separated list of IPs. Rightmost IP is the client IP. If the header appears multiple times, the last value will be used.
```

PR #392, opened 2025-10-13
Support lido modules

PedroCM96:
# Support for Lido Modules 

## Summary
This PR adds support for **all lido modules** by introducing a new configuration parameter, `lido_module_id`, with a default value of `1`.
Because the **Lido CS Module** exposes a different contract ABI, we've added a the corresponding ABI to the repository, and adapted this different logic accordingly.

## Motivation
- Enable the project to work with **any Lido module**, not just the default one.
- Ensure compatibility with **CSM**, which requires a **distinct ABI**.

## Code Updates

Changes in `crates/common/src/config/mux.rs` to:

- **Load the appropriate ABI** based on `lido_module_id`.
- **Handle differences** in ABI contract for **CSM**.
- **Preserve previous behavior** when `lido_module_id = 1 (default)`.

### DIFF
--- a/crates/common/src/types.rs
+++ b/crates/common/src/types.rs
@@ -47,7 +47,7 @@ pub enum Chain {
 pub enum MainnetLidoModule {
     Curated = 1,
     SimpleDVT = 2,
-    CommunityStaking = 3
+    CommunityStaking = 3,
 }
 
 pub enum HoleskyLidoModule {


crates/common/src/types.rs
```diff
@@ -44,6 +44,26 @@ pub enum Chain {
     },
 }
 
+pub enum MainnetLidoModule {
+    Curated = 1,
+    SimpleDVT = 2,
+    CommunityStaking = 3
```
jclapis:
```suggestion
    CommunityStaking = 3,
```

jclapis: `fmt-check` needs these commas here for CI to pass, Clippy doesn't care which is interesting; sorry about that!

PR #403, opened 2025-11-11
Add SSZ Support

jclapis:
This is an update to #372 which is now obsolete, and adds robust SSZ support to `get_header()` and `submit_block()`:
- Support for multiple ACCEPT types
- Support for dynamically converting SSZ requests when the relay only supports JSON
- Ability to translate transparently from the type returned by the relay back to the type requested by the client (e.g. SSZ (client request) -> JSON (relay request / response) -> SSZ (client response))


crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -305,43 +314,73 @@ async fn send_one_get_header(
...
+    // Check which types this request is for
+    let accept_types = get_accept_types(&req_config.headers).map_err(|e| {
+        PbsError::GeneralRequest(format!("error reading accept types: {e}").to_string())
+    })?;
+    let accepts_ssz = accept_types.contains(&EncodingType::Ssz);
+    let accepts_json = accept_types.contains(&EncodingType::Json);
+
+    // Send the header request
+    let mut start_request = Instant::now();
+    let config = RequestContext {
```
ltitanb: if we're just cloning the full `RequestContext` we should just do that as it's clearer

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -461,6 +528,56 @@ async fn send_one_get_header(
...
+                .with_label_values(&[TIMEOUT_ERROR_CODE_STR, GET_HEADER_ENDPOINT_TAG, &relay.id])
+                .inc();
+            return Err(err.into());
+        }
+    };
+
+    // Get the content type; this is only really useful for OK responses, and
+    // doesn't handle encoding types besides SSZ and JSON
+    let mut content_type: Option<EncodingType> = None;
+    if res.status() == StatusCode::OK &&
```
ltitanb: we can check if this is false and early return an error

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -363,15 +402,42 @@ async fn send_one_get_header(
...
-        Ok(parsed) => parsed,
-        Err(err) => {
-            return Err(PbsError::JsonDecode {
-                err,
-                raw: String::from_utf8_lossy(&response_bytes).into_owned(),
-            });
-        }
-    };
+    // Regenerate the header from the response
+    let get_header_response =
```
ltitanb: in general i think we should be "forgiving" as much as possible, eg. if the header is missing we try to parse as json

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -305,43 +314,73 @@ async fn send_one_get_header(
...
-    // the timestamp in the header is the consensus block time which is fixed,
-    // use the beginning of the request as proxy to make sure we use only the
-    // last one received
-    let start_request_time = utcnow_ms();
-    req_config.headers.insert(HEADER_START_TIME_UNIX_MS, HeaderValue::from(start_request_time));
+    let mut original_headers = req_config.headers.clone();
+
+    // Check which types this request is for
+    let accept_types = get_accept_types(&req_config.headers).map_err(|e| {
+        PbsError::GeneralRequest(format!("error reading accept types: {e}").to_string())
```
ltitanb: similar here, i think we can assume json if there's any issue

crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -269,6 +326,55 @@ async fn send_submit_block(
     Ok(Some(block_response))
 }
 
+async fn send_submit_block_impl(
```
ltitanb: similar here, i think we should move some of the logic here and only return a parsed block. So we submit and retry if there was an issue

ltitanb (changes requested):

crates/common/src/pbs/types/mod.rs
```diff
@@ -54,6 +64,17 @@ pub struct GetHeaderParams {
     pub pubkey: BlsPublicKey,
 }
 
+/// Which encoding types the original requester accepts in the response.
+/// As the builder spec adds more encoding types, this struct can be expanded.
+#[derive(Clone)]
+pub struct AcceptTypes {
```
ninaiiad: these seem to be unused, can delete? also we should have a clippy check that would flag things like that in CI

crates/common/src/config/pbs.rs
```diff
@@ -291,7 +322,7 @@ pub async fn load_pbs_config(config_path: Option<PathBuf>) -> Result<(PbsModuleC
     // Validate the muxes and build the lookup tables
     let (mux_lookup, registry_muxes) = match config.muxes {
         Some(muxes) => {
-            let (mux_lookup, registry_muxes) =
+            let (mux_lookup, registry_muxes): (_, _) =
```
ninaiiad: nit: can remove type annotation

crates/common/src/config/pbs.rs
```diff
@@ -373,7 +404,7 @@ pub async fn load_pbs_custom_config<T: DeserializeOwned>() -> Result<(PbsModuleC
     // Validate the muxes and build the lookup tables
     let (mux_lookup, registry_muxes) = match cb_config.muxes {
         Some(muxes) => {
-            let (mux_lookup, registry_muxes) = muxes
+            let (mux_lookup, registry_muxes): (_, _) = muxes
```
ninaiiad: here too

crates/common/src/utils.rs
```diff
@@ -517,6 +565,274 @@ pub fn get_user_agent_with_version(req_headers: &HeaderMap) -> eyre::Result<Head
...
+            // 0 means 'not acceptable'."
+            if let Some(q) = mt.get_param(mediatype::names::Q) &&
+                q.as_str().parse::<f32>().is_ok_and(|v| v <= 0.0)
+            {
+                continue;
+            }
+
+            let parsed = match mt.essence().to_string().as_str() {
+                APPLICATION_OCTET_STREAM => Some(EncodingType::Ssz),
+                APPLICATION_JSON | WILDCARD => Some(EncodingType::Json),
```
ninaiiad: is wildcard actually a valid json? should this match have two branches maybe: ssz <-> json?
there is a comment below that says "Assume missing Content-Type means JSON" too - ??

crates/common/src/utils.rs
```diff
@@ -517,6 +565,274 @@ pub fn get_user_agent_with_version(req_headers: &HeaderMap) -> eyre::Result<Head
...
+    ForkName::from_str(
+        req_headers
+            .get(CONSENSUS_VERSION_HEADER)
+            .and_then(|value| value.to_str().ok())
+            .unwrap_or(""),
+    )
+    .ok()
+}
+
+/// Enum for types that can be used to encode incoming request bodies or
```
ninaiiad: nit: redundant comment

crates/common/src/utils.rs
```diff
@@ -517,6 +565,274 @@ pub fn get_user_agent_with_version(req_headers: &HeaderMap) -> eyre::Result<Head
...
+/// outgoing response bodies
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum EncodingType {
+    /// Body is UTF-8 encoded as JSON
```
ninaiiad: so is this, just takes up space

crates/common/src/utils.rs
```diff
@@ -517,6 +565,274 @@ pub fn get_user_agent_with_version(req_headers: &HeaderMap) -> eyre::Result<Head
...
+    Json,
+
+    /// Body is raw bytes representing an SSZ object
+    Ssz,
+}
+
+impl EncodingType {
+    /// Get the content type string for the encoding type
```
ninaiiad: and this and all others

crates/common/src/utils.rs
```diff
@@ -517,6 +565,274 @@ pub fn get_user_agent_with_version(req_headers: &HeaderMap) -> eyre::Result<Head
...
+            }
+        };
+    }
+
+    Err(BodyDeserializeError::UnsupportedMediaType)
+}
+
+#[must_use]
+#[derive(Debug, Clone, Default)]
+pub struct RawRequest {
```
ninaiiad: why do we need a wrapper?

crates/common/src/utils.rs
```diff
@@ -517,6 +565,274 @@ pub fn get_user_agent_with_version(req_headers: &HeaderMap) -> eyre::Result<Head
...
+    pub body_bytes: Bytes,
+}
+
+impl<S> FromRequest<S> for RawRequest
+where
+    S: Send + Sync,
+{
+    type Rejection = AxumResponse;
+
+    async fn from_request(req: Request, _state: &S) -> Result<Self, Self::Rejection> {
```
ninaiiad: nit: _state -> state

crates/common/src/utils.rs
```diff
@@ -517,6 +565,274 @@ pub fn get_user_agent_with_version(req_headers: &HeaderMap) -> eyre::Result<Head
...
+            }
+        }
+    }
+}
+
+pub async fn deserialize_body(
+    headers: &HeaderMap,
+    body: Bytes,
+) -> Result<SignedBlindedBeaconBlock, BodyDeserializeError> {
+    if headers.contains_key(CONTENT_TYPE) {
```
ninaiiad: before this change looks like we tried deserializing body from json even if the content type header was omitted, maybe we should keep that?

crates/pbs/src/routes/get_header.rs
```diff
@@ -32,16 +37,103 @@ pub async fn handle_get_header<S: BuilderApiState, A: BuilderApi<S>>(
...
+                            res.headers_mut().insert(CONTENT_TYPE, content_type_header);
+                            info!("sending response as SSZ");
+                            return Ok(res);
+                        }
+
+                        // Handle JSON
+                        if accepts_json {
+                            Ok((StatusCode::OK, axum::Json(max_bid)).into_response())
+                        } else {
+                            // This shouldn't ever happen but the compiler needs it
```
ninaiiad: should we remove the if then? this should have three branches - 1) no encoding (which is handled somewhere above already so this is indeed unreachable), 2) ssz or 3) json

crates/pbs/src/routes/get_header.rs
```diff
@@ -32,16 +37,103 @@ pub async fn handle_get_header<S: BuilderApiState, A: BuilderApi<S>>(
...
+                            let mut res = max_bid.data.as_ssz_bytes().into_response();
+                            let consensus_version_header = match HeaderValue::from_str(
+                                &max_bid.version.to_string(),
+                            ) {
+                                Ok(consensus_version_header) => {
+                                    Ok::<HeaderValue, PbsClientError>(consensus_version_header)
+                                }
+                                Err(e) => {
+                                    if accepts_json {
+                                        info!("sending response as JSON");
```
ninaiiad: this too seems redundant

crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -1,33 +1,82 @@
...
+
+    /// How to validate the block returned by the relay
+    validation_mode: BlockValidationMode,
+
+    /// The accepted encoding types from the original request, ordered by
+    /// descending caller preference (q-value).
+    accepted_types: Vec<EncodingType>,
+}
+
+/// Used interally to provide info and context about a submit_block request and
```
ninaiiad: nit: redundant comments

crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -1,33 +1,82 @@
...
-        MAX_SIZE_SUBMIT_BLOCK_RESPONSE, SUBMIT_BLINDED_BLOCK_ENDPOINT_TAG, TIMEOUT_ERROR_CODE_STR,
-    },
-    metrics::{RELAY_LATENCY, RELAY_STATUS_CODE},
+    CompoundSubmitBlockResponse, LightSubmitBlockResponse, TIMEOUT_ERROR_CODE_STR,
+    constants::{MAX_SIZE_SUBMIT_BLOCK_RESPONSE, SUBMIT_BLINDED_BLOCK_ENDPOINT_TAG},
+    metrics::{RELAY_LATENCY, RELAY_STATUS_CODE, V2_FALLBACK_TO_V1},
     state::{BuilderApiState, PbsState},
 };
 
+/// Info about a proposal submission request.
```
ninaiiad: nit: also redundant

crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -1,33 +1,82 @@
...
+/// Sent from submit_block to the submit_block_with_timeout function.
+#[derive(Clone)]
+struct ProposalInfo {
+    /// The signed blinded block to submit
+    signed_blinded_block: Arc<SignedBlindedBeaconBlock>,
+
+    /// Common baseline of headers to send with each request
+    headers: Arc<HeaderMap>,
```
ninaiiad: I thought it should be relatively small, no? why Arc it?

crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -1,33 +1,82 @@
...
+
+    /// The version of the submit_block route being used
+    api_version: BuilderApiVersion,
+
+    /// How to validate the block returned by the relay
+    validation_mode: BlockValidationMode,
+
+    /// The accepted encoding types from the original request, ordered by
+    /// descending caller preference (q-value).
+    accepted_types: Vec<EncodingType>,
```
ninaiiad: this is max two values, right? can we actually get two accepted types and not one for a single request? if yes, can do an ArrayVec then, otherwise one value is enough

crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -190,96 +473,161 @@ async fn send_submit_block(
...
+        debug!(
+            relay_id = relay.id.as_ref(),
+            retry,
+            latency = ?request_latency,
+            "received 202 Accepted for v2 submit_block"
+        );
+
+        match code {
+            StatusCode::ACCEPTED => {
+                return Ok(SubmitBlockResponseInfo {
```
ninaiiad, line 538: maybe do an option instead? too expensive for a dummy value

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -285,54 +390,270 @@ async fn send_timed_get_header(
...
-struct ValidationContext {
-    skip_sigverify: bool,
-    min_bid_wei: U256,
-    extra_validation_enabled: bool,
-    parent_block: Arc<RwLock<Option<Block>>>,
+/// Send a get_header request and decode only the fork and bid value from the
+/// response, leaving the raw bytes intact for direct forwarding to the caller.
+/// Used in `HeaderValidationMode::None` where expensive crypto/structural
+/// checks are skipped.
+async fn send_get_header_light(
```
ninaiiad, line 570: looks like this does pretty much the same thing as the _full version, so the inside part of these can be extracted
can do single function returning the raw response_bytes + content_type + fork, with a single decode step driven by a flag for the full / light

crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -159,22 +221,243 @@ async fn submit_block_with_timeout(
...
+        version =% decoded_response.version,
+        "received unblinded block"
+    );
+
+    Ok(Some(decoded_response))
+}
+
+/// Send and lightly process a submit_block request, minimizing the amount of
+/// decoding and validation done
+async fn send_submit_block_light(
```
ninaiiad, line 363: same here - some duplicate code with the send_submit_block_full that can be extracted

crates/common/src/utils.rs
```diff
@@ -29,11 +48,40 @@ use tracing_subscriber::{
...
 
+pub const APPLICATION_JSON: &str = "application/json";
+pub const APPLICATION_OCTET_STREAM: &str = "application/octet-stream";
+pub const WILDCARD: &str = "*/*";
+
 const MILLIS_PER_SECOND: u64 = 1_000;
+pub const CONSENSUS_VERSION_HEADER: &str = "Eth-Consensus-Version";
+
+lazy_static! {
+    static ref SSZ_VALUE_OFFSETS_BY_FORK: HashMap<ForkName, usize> = {
```
ninaiiad: don't think we need a hashmap for this, performance-wise inlining the get_ssz_value_offset_for_fork should do better

ninaiiad (changes requested):

ninaiiad (approved):

Issue #382, opened 2025-09-29
Feature request: Auto-Reload Configuration on Detection of File and SSV loader Detection Changes


meta-duck:
# Auto-Reload Configuration on Detection of File and SSV loader Detection Changes

## Description
I would like cb able to automatically reload the configuration when it detects changes in either the configuration file (`cb-config.toml`) and the SSV node operator status (e.g., when a new validator key becomes active or added in the SSV network). This would streamline the process for newly deposited/activated validator keys without requiring manual hot reloads via a POST request to the `/reload` endpoint.

## Current Behavior
- A manual hot reload (via a POST request to `/reload`) is required to apply the updated configuration for the new key to enable the mux option. If no hot reload is performed, the key uses the default relay list (no mux).
- cb dose not detect changes in the SSV node operator change automatically.

PR #409, opened 2025-11-19
PBS: Reload the config file when it changes

jclapis:
This closes the first request in #382 by automatically reloading the config for the PBS service as though `/reload` were called whenever the config file changes. There are a few ways to do this, but I settled on propagating the path from the config loader instead of making it a field within the `PbsConfig`, because it didn't feel like a good fit for that.

Note that the Signer service doesn't do hot-reloading, because it's a much different setup that will require more planning. That might be OBE during the config revamp anyway so I'm leaving it out for now.

### DIFF
--- a/crates/pbs/src/service.rs
+++ b/crates/pbs/src/service.rs
@@ -68,12 +68,20 @@ impl PbsService {
             let config_path_for_watcher = config_path.clone();
             watcher = RecommendedWatcher::new(
                 move |result: Result<Event, Error>| {
-                    let event = result.unwrap();
-                    if !event.kind.is_modify() {
-                        return;
+                    match result {
+                        Err(e) => {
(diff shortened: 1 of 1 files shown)


crates/pbs/src/service.rs
```diff
@@ -59,6 +61,38 @@ impl PbsService {
             bail!("PBS server failed to start. Are the relays properly configured?");
         }
 
+        // Set up the filesystem watcher for the config file
+        let mut watcher: RecommendedWatcher;
+        if config_path.to_str() != Some("") {
+            let state_for_watcher = state.clone();
```
ltitanb: would add a couple of logs:
- that we start the file watcher
- when trying to reload the config

jclapis: Sure, done in 8d77531.

PR #223, opened 2025-01-08
feat(mux): support SSV network

ManuelBilbao:
Add support for SSV network in the PBS mux

Related: #15

PR #415, opened 2025-12-05
ssv-network: SSV-node API support

iurii-ssv:
This PR should allow a commit-boost operator to use the API of their own SSV node (instead of relying on a 3rd party API that is currently available at api.ssv.network).

The SSV node API looks like this (it is different from the API-endpoint exposed by api.ssv.network service):
- request example:
```
curl --request GET \
  --url {ssv_node_base_url}/v1/validators \
  --header 'Content-Type: application/json' \
  --data '{
  "operators": [1]
}'
```
- response example:
```
{
  "data": [
    {
      "public_key": "aa370f6250d421d00437b9900407a7ad93b041aeb7259d99b55ab8b163277746680e93e841f87350737bceee46aa104d",
      "index": "1311498",
      "status": "active_ongoing",
      "activation_epoch": "273156",
      "exit_epoch": "18446744073709551615",
      "owner": "5e33db0b37622f7e6b2f0654aa7b985d854ea9cb",
      "committee": [
        1,
        2,
        3,
        4
      ],
      "quorum": 0,
      "partial_quorum": 0,
      "graffiti": "",
      "liquidated": false
    }
  ]
}
```

See also the [original implementation PR (the old way)](https://github.com/Commit-Boost/commit-boost-client/pull/223)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ cb-metrics = { path = "crates/metrics" }
 cb-pbs = { path = "crates/pbs" }
 cb-signer = { path = "crates/signer" }
 cipher = "0.4"
-clap = { version = "4.5.4", features = ["derive", "env"] }
+clap = { version = "4.5.48", features = ["derive", "env"] }
 color-eyre = "0.6.3"
 ctr = "0.9.2"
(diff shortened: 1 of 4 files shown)


iurii-ssv (commented):
Awesome! Looks good to me.

ltitanb (approved):

PR #426, opened 2026-02-15
feat(pbs): Log auction participants with bid value and delta at INFO level

selfuryon:
We have auction logs in Vouch which I really like. This log shows bid auction and who won it with which value, so I also want to add that type of logs in PBS module too


JasonVranek:
closed by #443

PR #438, opened 2026-03-19
Sigp audit fixes

JasonVranek:
Implements the recommended changes to address issues from the Sigma Prime audit report.


crates/common/src/config/utils.rs
```diff
@@ -101,6 +101,7 @@ mod tests {
     use super::*;
     use crate::utils::TestRandomSeed;
 
+    /// TODO: This was only used by the old JWT loader, can it be removed now?
```
JasonVranek: remove

crates/common/src/config/signer.rs
```diff
@@ -240,3 +366,553 @@ impl StartSignerConfig {
...
+    config: &CommitBoostConfig,
+    jwt_secrets: &HashMap<ModuleId, String>,
+) -> Result<HashMap<ModuleId, ModuleSigningConfig>> {
+    let mut mod_signing_configs = HashMap::new();
+    let modules = config.modules.as_ref().ok_or_eyre("No modules defined in the config")?;
+
+    let mut seen_jwt_secrets = HashMap::new();
+    let mut seen_signing_ids = HashMap::new();
+    for module in modules {
+        // Validate the module ID
```
ninaiiad: nit: same as above. most other comments in this function are doing much either

ninaiiad (approved):

ninaiiad (approved):

ltitanb (dismissed):

crates/common/Cargo.toml
```diff
@@ -51,3 +52,7 @@ tree_hash.workspace = true
 tree_hash_derive.workspace = true
 unicode-normalization.workspace = true
 url.workspace = true
+reqwest-eventsource = "=0.5.0"
```
ManuelBilbao: This should be in the root `Cargo.toml` and use `.workspace = true` here

crates/signer/src/service.rs
```diff
@@ -354,40 +620,81 @@ async fn handle_generate_proxy(
...
+        let mut jwt_configs = state.jwts.write();
+        for (module_id, jwt_secret) in jwt_secrets {
+            if let Some(cfg) = jwt_configs.get_mut(&module_id) {
+                cfg.jwt_secret = jwt_secret;
+            } else {
+                let error_message = format!(
+                    "Module {module_id} signing ID not found in commit-boost config, cannot reload"
+                );
+                error!(event = "reload", module_id = %module_id, error = %error_message);
+                return Err(SignerModuleError::RequestError(error_message));
```
ninaiiad, line 682: if we're able to return midway through the loop it's strange that we're also updating the configs in place because we may exit and return an error to the client having half-updated the configs
It doesn't look like an actual vulnerability, but it's not clean and it's not obvious to the client I think

if you think it's better to be able to do this, then I think we should be able to communicate to the client which modules have been updated, also maybe save the error and continue updating other modules afterwards
though i would prefer to check first, update later only if all are ok

crates/signer/src/service.rs
```diff
@@ -108,124 +125,255 @@ impl SigningService {
...
-        // Rate limit the request
-        let remaining = state.jwt_auth_fail_timeout.saturating_sub(elapsed);
-        warn!(
-            "Client {client_ip} is rate limited for {remaining:?} more seconds due to JWT auth failures"
+    // If the last failure was more than the timeout ago, remove this entry so it's
+    // eligible again
+    if elapsed > state.jwt_auth_fail_timeout {
+        drop(failures);
+        debug!("Removing {client_ip} from JWT auth failure list");
+        state.jwt_auth_failures.write().remove(client_ip);
```
ninaiiad: this is a concurrency issue - we can't expect the state to have stayed the same after we've dropped the read lock above
there could've been another failure we missed in between the dropping of the read lock and the acquiring of the write lock where a new failure was added
not likely a big vulnerability either, but a bug nevertheless

ninaiiad (dismissed):

docs/docs/get_started/configuration.md
```diff
@@ -452,6 +518,15 @@ Commit-Boost supports hot-reloading the configuration file. This means that you
 docker compose -f cb.docker-compose.yml exec cb_signer curl -X POST http://localhost:20000/reload
 ```
 
+### Signer module reload
+
+The signer module takes 2 optional parameters in the JSON body:
+
+- `jwt_secrets`: a string with a comma-separated list of `<MODULE_ID>=<JWT_SECRET>` for all modules.
```
ManuelBilbao: It might be worth mentioning that this list will replace existing one, so old values should be included if they are intended to be kept

crates/common/src/signer/schemes/ecdsa.rs
```diff
@@ -86,58 +86,113 @@ impl EcdsaSigner {
...
+                            module_signing_id: *module_signing_id,
+                            nonce: *nonce,
+                            chain_id: chain.id(),
+                        }
+                        .tree_hash_root();
+                        types::SigningData { object_root, signing_domain }.tree_hash_root()
+                    }
+                    None => types::SigningData { object_root: *object_root, signing_domain }
+                        .tree_hash_root(),
+                };
```
ManuelBilbao: Should we use `signature::compute_prop_commit_signing_root()` instead?

```suggestion
                let signing_root = crate::signature::compute_prop_commit_signing_root(
                    chain,
                    object_root,
                    signature_request_info,
                    &B32::from(COMMIT_BOOST_DOMAIN),
                );
```

crates/signer/src/manager/dirk.rs
```diff
@@ -233,13 +245,28 @@ impl DirkManager {
...
+                    data: *object_root,
+                    module_signing_id: *module_signing_id,
+                    nonce: *nonce,
+                    chain_id: self.chain.id(),
+                }
+                .tree_hash_root()
+                .to_vec()
+            }
+            None => object_root.to_vec(),
+        };
```
ManuelBilbao, line 265: Maybe same thing here about using `compute_prop_commit_signing_root`

crates/signer/src/service.rs
```diff
@@ -108,124 +125,255 @@ impl SigningService {
...
             jwt_auth_fail_limit =? state.jwt_auth_fail_limit,
             jwt_auth_fail_timeout =? state.jwt_auth_fail_timeout,
+            reverse_proxy =% state.reverse_proxy,
             "Starting signing service"
         );
 
         SigningService::init_metrics(config.chain)?;
 
-        let app = axum::Router::new()
-            .route(REQUEST_SIGNATURE_PATH, post(handle_request_signature))
```
ManuelBilbao, line 117: Should we keep the path as deprecated to avoid sudden breaking changes?

docs/docs/get_started/configuration.md
JasonVranek: apply_reload() behavior was recently changed to allow partial reloads, so if you omit a jwt secret it keeps the old value instead of silently dropping it

ManuelBilbao: Is there a way to add/remove modules?

crates/signer/src/manager/dirk.rs
JasonVranek: compute_prop_commit_signing_root mixes in the domain but then the dirk SignRequest accepts the domain as a param so we don't want double domain mixing

crates/signer/src/service.rs
```diff
@@ -648,37 +650,49 @@ async fn handle_reload(
...
     state: SigningState,
     request: ReloadRequest,
     new_manager: SigningManager,
+    new_signing_configs: HashMap<ModuleId, ModuleSigningConfig>,
+    new_admin_secret: String,
 ) -> Result<StatusCode, SignerModuleError> {
-    // Update the JWT configs if provided in the request. Only the provided
-    // modules are updated; omitted modules keep their existing secrets.
+    // Build the new JWT map from config, then apply any body overrides.
+    let mut new_jwts = new_signing_configs;
```
ninaiiad, line 673: you can make the function argument mutable, then we don't need a separate variable

crates/signer/src/service.rs
```diff
@@ -640,6 +638,10 @@ async fn handle_reload(
         }
     };
 
+    // Extract the fields we need before start_manager consumes the config
+    let new_signing_configs = config.mod_signing_configs.clone();
```
ninaiiad, line 642: looks like the start_manager function doesn't access these fields, so we could move the ones it needs into it and then move these two into the apply_reload() instead of cloning

ltitanb (approved):

ManuelBilbao (approved):

PR #464, opened 2026-04-29
Feature/commit boost default image

JasonVranek:
We recently updated the release process on main, but this targeted main after the last hotfix which had temporarily reverted code. Specifically it undid #425 which unified the `pbs`, `signer`, and `cli` binaries into just `commit-boost`. Since the new release process didn't include these changes, this PR captures them. 

Specifically, we're introducing a new `commit-boost` Docker image to the release process that defaults to use the `pbs` command. This is the new de facto image used throughout the project, including the Justfile. However, to remain backwards compatible, each release will continue to ship a `pbs:latest` and `signer:latest` image, which are simply clones of the `commit-boost` image but hardwired to use their `pbs` and `signer` commands respectively.

### DIFF
--- a/crates/common/src/config/constants.rs
+++ b/crates/common/src/config/constants.rs
@@ -16,7 +16,7 @@ pub const LOGS_DIR_DEFAULT: &str = "/var/logs/commit-boost";
 
 ///////////////////////// PBS /////////////////////////
 
-pub const PBS_IMAGE_DEFAULT: &str = "ghcr.io/commit-boost/pbs:latest";
+pub const PBS_IMAGE_DEFAULT: &str = "ghcr.io/commit-boost/commit-boost:latest";
 pub const PBS_SERVICE_NAME: &str = "pbs";
 
 /// Where to receive BuilderAPI calls from beacon node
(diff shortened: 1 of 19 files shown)


crates/common/src/config/constants.rs
```diff
@@ -16,7 +16,7 @@ pub const LOGS_DIR_DEFAULT: &str = "/var/logs/commit-boost";
 
 ///////////////////////// PBS /////////////////////////
 
-pub const PBS_IMAGE_DEFAULT: &str = "ghcr.io/commit-boost/pbs:latest";
+pub const PBS_IMAGE_DEFAULT: &str = "ghcr.io/commit-boost/commit-boost:latest";
```
ManuelBilbao: nit: Maybe just replace with something like `DOCKER_IMAGE_DEFAULT` for both services

provisioning/k8s/commit-boost/values.yaml
```diff
@@ -110,7 +110,7 @@ commitBoost:
   pbs:
     enable: true
     image:
-      repository: ghcr.io/commit-boost/pbs
+      repository: ghcr.io/commit-boost/commit-boost
```
ManuelBilbao, line 113: Should we also explicit the `pbs` command in the deployment manifest?

ManuelBilbao (approved):

PR #465, opened 2026-05-08
feat: add Stader validator registry support

PedroCM96:
Add support for loading validator keys from Stader registries.

A new `stader` registry is introduced with explicit pool selection for Stader's permissioned or permissionless node registries.

This implementation follows the same high-level structure used for the Lido implementation: resolve the registry contract, fetch the total number of keys and load those keys in batches through the shared registry key collection logic.


JasonVranek (dismissed):

ltitanb (approved):

ltitanb (dismissed):

ltitanb (dismissed):

ManuelBilbao (changes requested):
Please update the loader reference in `config.example.toml`

https://github.com/Commit-Boost/commit-boost-client/blob/635384f4c9d43289ac1eb2ad09a09c076de17eeb/config.example.toml#L139-L160

ManuelBilbao (approved):

ltitanb (approved):

PR #467, opened 2026-05-20
Rewrite get_header relay communication to support SSZ and JSON

JasonVranek:
## Summary

Rewrites get_header relay communication to support both SSZ and JSON content negotiation.

## What's in this PR

- SSZ and JSON encoding/decoding for relay get_header responses
- Content-Type and Eth-Consensus-Version header parsing on relay responses
- Accept header forwarding to relays with q-value ordering
- MIME parameter tolerance on relay response Content-Type
- Fork-aware response decoding for all supported forks
- Updated mock relay and mock validator with SSZ support
- Dynamic port allocation in tests via `get_free_listener` (eliminates flaky CI from port collisions)
- Get_header integration tests covering both encodings
- Per-fork JSON fixture files for test data

## Review notes

The route handler always takes the standard validation path in this PR. Configurable validation modes come in PR 4.

The test infrastructure changes (mock relay SSZ support, dynamic ports) are shared by both get_header and submit_block tests.

Part 2 of 4: SSZ types → get_header → submit_block → validation bypassing

## Attribution

This work builds on the SSZ builder flow originally implemented by @eserilev 
in #252 and ported to the current codebase by @jclapis in #403.

Co-authored-by: Eitan Seri-Levi <eserilev@ucsc.edu>
Co-authored-by: Joe Clapis <jclapis@outlook.com>


crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -283,54 +347,175 @@ async fn send_timed_get_header(
...
+        Some(info) => info,
+        None => {
+            return Ok((start_request_time, None));
+        }
+    };
+
+    // Decode the response
+    let get_header_response = decode_by_encoding(&info, decode_json_payload, decode_ssz_payload)?;
+
+    // Log and return
```
ninaiiad: redundant

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -283,54 +347,175 @@ async fn send_timed_get_header(
...
+    let (start_request_time, info) =
+        send_get_header_impl(relay, url, timeout_left_ms, headers).await?;
+    let info = match info {
+        Some(info) => info,
+        None => {
+            return Ok((start_request_time, None));
+        }
+    };
+
+    // Decode the response
```
ninaiiad: this too

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -283,54 +347,175 @@ async fn send_timed_get_header(
...
-    url: Url,
-    timeout_ms: u64,
+/// Send and decode a full get_header response, with all of the fields.
+async fn send_get_header_full(
+    relay: &RelayClient,
+    url: Arc<Url>,
+    timeout_left_ms: u64,
     headers: HeaderMap,
+) -> Result<(u64, Option<GetHeaderResponse>), PbsError> {
+    // Send the request
```
ninaiiad: this too

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -283,54 +347,175 @@ async fn send_timed_get_header(
...
+    let validation = &request_info.validation;
+    validate_header_data(
+        &header_data,
+        chain,
+        params.parent_hash,
+        validation.min_bid_wei,
+        params.slot,
+    )?;
+
+    // Validate the relay signature
```
ninaiiad: and this

tests/src/mock_relay.rs
```diff
@@ -184,17 +304,61 @@ async fn handle_register_validator(
...
+
+    // Error out if the request content type is not supported
+    let content_type = get_content_type(&headers);
+    if !state.supported_content_types.contains(&content_type) {
+        return (StatusCode::UNSUPPORTED_MEDIA_TYPE, "Unsupported content type".to_string())
+            .into_response();
+    };
+
+    let data = if state.large_body() {
+        vec![1u8; 1 + MAX_SIZE_SUBMIT_BLOCK_RESPONSE]
```
ninaiiad, line 351: why do we need this?

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -95,22 +149,40 @@ pub async fn get_header<S: BuilderApiState>(
...
+    // SSZ first (wire-efficient), JSON fallback.
+    let accept_types = OUTBOUND_ACCEPT.to_string();
+    send_headers.insert(
+        ACCEPT,
+        HeaderValue::from_str(&accept_types).map_err(|e| {
+            AcceptedEncodingsError::InvalidEncoding {
+                error_msg: (format!("invalid accept header: {e}")).to_string(),
+            }
+        })?,
+    );
```
ninaiiad: send_headers.insert(ACCEPT, HeaderValue::from_static(OUTBOUND_ACCEPT));

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -364,101 +552,42 @@ async fn send_one_get_header(
...
-                )?;
-            }
-        }
+/// Decode a JSON-encoded get_header response
+fn decode_json_payload(response_bytes: &[u8]) -> Result<GetHeaderResponse, PbsError> {
+    match serde_json::from_slice::<GetHeaderResponse>(response_bytes) {
+        Ok(parsed) => Ok(parsed),
+        Err(err) => Err(PbsError::JsonDecode {
+            err,
+            raw: String::from_utf8_lossy(response_bytes).into_owned(),
```
ninaiiad, line 577: do you need the into_owned here?

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -364,101 +552,42 @@ async fn send_one_get_header(
...
-                "parent block not found, skipping extra validation"
-            );
+/// Decode an SSZ-encoded get_header response
+fn decode_ssz_payload(
+    response_bytes: &[u8],
+    fork: ForkName,
+) -> Result<GetHeaderResponse, PbsError> {
+    let data = SignedBuilderBid::from_ssz_bytes_by_fork(response_bytes, fork).map_err(|e| {
+        PbsError::SSZDecode {
+            err: (format!("error decoding relay payload: {e:?}")).to_string(),
```
ninaiiad: don't need the to_string() call

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -177,15 +249,13 @@ async fn fetch_parent_block(
...
     relay: RelayClient,
-    chain: Chain,
-    headers: HeaderMap,
     ms_into_slot: u64,
     mut timeout_left_ms: u64,
-    validation: ValidationContext,
 ) -> Result<Option<GetHeaderResponse>, PbsError> {
-    let url = relay.get_header_url(params.slot, &params.parent_hash, &params.pubkey)?;
+    let params = &request_info.params;
+    let url = Arc::new(relay.get_header_url(params.slot, &params.parent_hash, &params.pubkey)?);
```
ninaiiad: claude:
> let url = Arc::new(relay.get_header_url(...)?);                        // :258
  // ...spawned: send_one_get_header(request_info.clone(), relay.clone(), url.clone(), ...)
  // ...inside: .get(url.as_ref().clone())                              // :516 — full Url clone anyway
reqwest's IntoUrl has no &Url impl, so .get() needs an owned Url regardless — hence url.as_ref().clone() clones the Url (a String alloc) on every request. The Arc therefore buys nothing: you still do one full Url clone per HTTP send, plus an Arc allocation up front and an atomic bump per spawn.
Pass an owned Url instead (url: Url): move it into .get(url) on the single-request path (zero clones), and url.clone() per spawn only in the timing-games loop (same count of Url clones as today, minus all the Arc overhead). Strictly fewer allocations in the common path.

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -39,6 +46,53 @@ use crate::{
...
+/// Info about an incoming get_header request.
+/// Sent from get_header to each send_timed_get_header call.
+#[derive(Clone)]
+struct RequestInfo {
+    /// The blockchain parameters of the get_header request (what slot it's for,
+    /// which pubkey is requesting it, etc)
+    params: GetHeaderParams,
+
+    /// Common baseline of headers to send with each request
+    headers: Arc<HeaderMap>,
```
ninaiiad: why does this need an Arc? we never clone it as an Arc, only the full RequestInfo struct which clones the hash map

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -283,54 +347,175 @@ async fn send_timed_get_header(
...
-struct ValidationContext {
-    skip_sigverify: bool,
-    min_bid_wei: U256,
-    extra_validation_enabled: bool,
-    parent_block: Arc<RwLock<Option<Block>>>,
+/// Dispatch a get_header response to the appropriate decoder based on the
+/// negotiated content-type. SSZ requires a fork header; its absence is a
+/// protocol violation reported as `PbsError::RelayResponse`. Callers supply
+/// the format-specific decoders, keeping the encoding branch in one place.
+fn decode_by_encoding<T>(
```
ninaiiad: this is only used once I think, not sure we need the generic parameter

crates/common/src/wire.rs
ninaiiad: you don't need to allocate a string here to do a match:
```rust
fn essence_encoding(mt: &MediaType) -> Option<EncodingType> {
    if mt.suffix.is_some() || mt.ty != names::APPLICATION {
        return None;
    }
    if mt.subty == names::OCTET_STREAM {
        Some(EncodingType::Ssz)
    } else if mt.subty == names::JSON {
        Some(EncodingType::Json)
    } else {
        None
    }
}
```

ninaiiad, line 375: same here

crates/pbs/src/mev_boost/get_header.rs
JasonVranek: `from_utf8_lossy` returns `Cow<'_, str>`

tests/src/mock_relay.rs
JasonVranek: we have a `test_submit_block_too_large`, the next PR that targets submit_block with ssz cleans up a lot of this logic

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -283,54 +346,169 @@ async fn send_timed_get_header(
...
-    req_config.headers.insert(HEADER_TIMEOUT_MS, HeaderValue::from(req_config.timeout_ms));
+    headers.insert(HEADER_TIMEOUT_MS, HeaderValue::from(timeout_left_ms));
 
     let start_request = Instant::now();
     let res = match relay
         .client
-        .get(req_config.url)
-        .timeout(Duration::from_millis(req_config.timeout_ms))
-        .headers(req_config.headers)
+        .get(url.as_ref())
```
ninaiiad: >   Url's only AsRef impl is AsRef<str>, so url.as_ref() resolves to &str, and reqwest's IntoUrlSealed for &str does Url::parse(self) (reqwest-0.12.28 into_url.rs:46-49). So every request now re-parses the URL string, which is strictly more expensive than the Url::clone it replaced

let's do `.get(url)`

crates/common/src/wire.rs
```diff
@@ -243,14 +255,30 @@ pub fn get_accept_types(req_headers: &HeaderMap) -> eyre::Result<AcceptedEncodin
...
+    }
+
+    match () {
+        _ if mt.ty == names::_STAR && mt.subty == names::_STAR => Some(NO_PREFERENCE_DEFAULT),
+        _ if mt.ty == names::APPLICATION && mt.subty == names::OCTET_STREAM => {
+            Some(EncodingType::Ssz)
+        }
+        _ if mt.ty == names::APPLICATION && mt.subty == names::JSON => Some(EncodingType::Json),
+        _ if mt.ty == names::APPLICATION && mt.subty == names::_STAR => Some(NO_PREFERENCE_DEFAULT),
+        _ => None,
```
ninaiiad, line 274: we used to reject Content-Type: */*, now we don't

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -95,22 +149,39 @@ pub async fn get_header<S: BuilderApiState>(
     let mut send_headers = HeaderMap::new();
     send_headers.insert(USER_AGENT, get_user_agent_with_version(&req_headers)?);
 
+    // Create the Accept headers for requests
+    // Use the documented, deterministic preference:
+    // SSZ first (wire-efficient), JSON fallback.
+    send_headers.insert(
+        ACCEPT,
+        HeaderValue::from_str(OUTBOUND_ACCEPT).map_err(|e| {
```
ninaiiad: let's do `from_static()`

crates/common/src/wire.rs
JasonVranek: added a regression tests for `Accept: /`, and `Content-Type: /`

- for `Accept: /` we do error so the essence_encoding() doesn't seem to have changed it
- for `Content-Type: /` we actually use `get_content_type()` which defaults on failure to JSON

ninaiiad: sorry, I meant `*/*` but the stars did not show up in the final message

ninaiiad (approved):

crates/common/src/wire.rs
JasonVranek: opting to keep mapping `*/*` and `application/*` to JSON so get_header falls back to something beacon nodes understand if they accidentally send `*/*`

crates/pbs/src/error.rs
```diff
@@ -22,17 +27,19 @@ impl PbsClientError {
             PbsClientError::NoPayload => StatusCode::BAD_GATEWAY,
             PbsClientError::Internal => StatusCode::INTERNAL_SERVER_ERROR,
             PbsClientError::DecodeError(_) => StatusCode::BAD_REQUEST,
+            PbsClientError::HeaderError(_) => StatusCode::BAD_REQUEST,
```
ManuelBilbao: There's a test saying that it can return 406. I'd change either of them to match

https://github.com/Commit-Boost/commit-boost-client/blob/a52b531bc16102d0d95ed28348d64707ce24bc8a/crates/common/src/wire.rs#L612-L614

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -95,22 +149,32 @@ pub async fn get_header<S: BuilderApiState>(
     let mut send_headers = HeaderMap::new();
     send_headers.insert(USER_AGENT, get_user_agent_with_version(&req_headers)?);
 
+    // Create the Accept headers for requests
+    // Use the documented, deterministic preference:
+    // SSZ first (wire-efficient), JSON fallback.
+    send_headers.insert(ACCEPT, HeaderValue::from_static(OUTBOUND_ACCEPT));
```
ManuelBilbao: To make sure, in the description you mention that headers are forwarded to the relay but here we are hardcoding the `accept`. Is this behaviour expected?

ManuelBilbao: Furthermore, we have this function that seems to do the job, but it's never used anywhere

https://github.com/Commit-Boost/commit-boost-client/blob/a52b531bc16102d0d95ed28348d64707ce24bc8a/crates/common/src/wire.rs#L300-L304

tests/src/utils.rs
```diff
@@ -27,6 +27,18 @@ pub fn get_local_address(port: u16) -> String {
...
 }
 
+/// Bind to port 0 and let the OS assign an unused ephemeral port.
+///
+/// The returned listener keeps the port reserved. Pass it to
+/// [`start_mock_relay_service_with_listener`] so the socket is never released
+/// between allocation and use (zero TOCTOU race). For servers that bind
+/// internally (e.g. `PbsService::run`), read the port with
+/// `listener.local_addr().unwrap().port()`, then `drop` the listener
+/// immediately before starting the server.
```
ltitanb: we should try to avoid these llm comments to leak into the codebase, we should have some guidelines on an AGENTS.md

crates/pbs/src/mev_boost/status.rs
```diff
@@ -73,8 +73,8 @@ async fn send_relay_check(relay: &RelayClient, headers: HeaderMap) -> Result<(),
     let code = res.status();
     RELAY_STATUS_CODE.with_label_values(&[code.as_str(), STATUS_ENDPOINT_TAG, &relay.id]).inc();
 
-    safe_read_http_response(res, MAX_SIZE_DEFAULT).await.inspect_err(|e| {
-        error!(relay_id = relay.id.as_ref(), %e, "status failed");
+    safe_read_http_response(res, MAX_SIZE_DEFAULT).await.inspect_err(|err| {
+        error!(relay_id = relay.id.as_ref(), %err, "status failed");
```
ltitanb, line 77: im in favour of this cleanup to be consistent throughout the codebase, just need to be careful when changing logs as i assume some people are parsing them for alerts/analytics etc

crates/common/src/wire.rs
```diff
@@ -198,16 +210,22 @@ impl IntoIterator for AcceptedEncodings {
...
     let mut fallback: Option<EncodingType> = None;
     let mut saw_any = false;
     let mut had_supported = false;
     for header in req_headers.get_all(ACCEPT).iter() {
-        let accept = Accept::from_str(header.to_str()?)
-            .map_err(|e| eyre::eyre!("invalid accept header: {e}"))?;
+        let accept_str = header.to_str().map_err(AcceptedEncodingsError::InvalidString)?;
+        let accept =
+            Accept::from_str(accept_str).map_err(|e| AcceptedEncodingsError::InvalidEncoding {
+                error_msg: (format!("invalid accept header: {e}")).to_string(),
```
ltitanb: in the error msg above you already have `[error("invalid accept header: {error_msg}")]` so this would be printed twice. I would just pass the error, also saves the string alloc

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -39,6 +46,53 @@ use crate::{
     utils::check_gas_limit,
 };
 
+/// Info about an incoming get_header request.
+/// Sent from get_header to each send_timed_get_header call.
+#[derive(Clone)]
+struct RequestInfo {
+    /// The blockchain parameters of the get_header request (what slot it's for,
+    /// which pubkey is requesting it, etc)
```
ltitanb: could we trim down on LLM comments here and below

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -283,54 +351,169 @@ async fn send_timed_get_header(
...
-    parent_block: Arc<RwLock<Option<Block>>>,
+/// Dispatch a get_header response to the appropriate decoder based on the
+/// negotiated content-type. SSZ requires a fork header; its absence is a
+/// protocol violation reported as `PbsError::RelayResponse`. Callers supply
+/// the format-specific decoders, keeping the encoding branch in one place.
+fn decode_by_encoding(
+    info: &GetHeaderResponseInfo,
+    on_json: impl FnOnce(&[u8]) -> Result<GetHeaderResponse, PbsError>,
+    on_ssz: impl FnOnce(&[u8], ForkName) -> Result<GetHeaderResponse, PbsError>,
+) -> Result<GetHeaderResponse, PbsError> {
```
ltitanb, line 475: this pattern makes it quite hard to follow / read. Why do we need closures if there's just one call site?

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -665,4 +801,43 @@ mod tests {
...
+
+            // Extract the bid value from the SSZ
+            let encoded = decoded.data.as_ssz_bytes();
+            let bid_value = get_bid_value_from_signed_builder_bid_ssz(&encoded, fork_name)
+                .expect("failed to extract bid value from SSZ");
+
+            // Compare to the original value
+            println!("Testing fork: {}", fork_name);
+            println!("Original value: {}", decoded.value());
+            println!("Extracted value: {}", bid_value);
```
ltitanb: assume these were for debugging but can remove it now if the test passes?

ltitanb (changes requested):

crates/pbs/src/mev_boost/get_header.rs
JasonVranek: In #469 when we reintroduce validation bypassing there are 2 callsites and when we originally reviewed that code we opted to split it into decode_by_encoding for better readability

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -95,22 +147,43 @@ pub async fn get_header<S: BuilderApiState>(
     let mut send_headers = HeaderMap::new();
     send_headers.insert(USER_AGENT, get_user_agent_with_version(&req_headers)?);
 
+    // Forward the caller's Accept preference to the relay so the relay
+    // returns data in the format the BN expects, avoiding decode→re-encode.
+    // Always offer both encodings as fallback so a format-limited relay
+    // still returns a bid (PBS converts if needed).
+    let caller_accept = get_accept_types(&req_headers).inspect_err(|err| {
```
ManuelBilbao, line 154: `get_accept_types` is also called in the caller function (`routes::get_header::get_header`). I'd try to avoid the duplicate call if possible

crates/pbs/src/routes/get_header.rs
```diff
@@ -33,16 +37,51 @@ pub async fn handle_get_header<S: BuilderApiState, A: BuilderApi<S>>(
...
+    // Honor caller q-value preference: pick the highest-priority encoding that
+    // we can actually produce. Server preference for tiebreaks is SSZ first.
+    let response_encoding = accept_types.preferred(&[EncodingType::Ssz, EncodingType::Json]);
 
     info!(ua, ms_into_slot, "new request");
 
     match A::get_header(params, req_headers, state).await {
         Ok(res) => {
             if let Some(max_bid) = res {
+                BEACON_NODE_STATUS.with_label_values(&["200", GET_HEADER_ENDPOINT_TAG]).inc();
```
ManuelBilbao: If the match goes through `None` branch, the response code is not 200

crates/pbs/src/mev_boost/get_header.rs
JasonVranek: we can more easily dedup this what we tackle the planned removal of the BuilderApi trait

ltitanb (approved):

ManuelBilbao (approved):

PR #468, opened 2026-05-20
Rewrite submit_block relay communication to support SSZ and JSON

JasonVranek:
## Summary

Rewrites submit_block relay communication to support SSZ and JSON content negotiation.

## What's in this PR

- SSZ-first request encoding with automatic JSON fallback on 406/415 (per builder-spec)
- Content-Type and Eth-Consensus-Version header handling for submit_block
- Fork-aware SSZ decoding for relay responses
- v2→v1 fallback now forwards the v1 payload to the beacon node instead of returning an empty 202 (prevents silent block loss)
- `V2_FALLBACK_TO_V1` metric counter for observability
- Submit_block integration tests covering SSZ/JSON encoding, v1/v2 paths, and fallback behavior

## Review notes

The v2→v1 fallback behavior is an intentional change from the previous code which silently discarded the v1 payload. The beacon node needs the unblinded execution payload to broadcast the block (returning an empty 202 on fallback would cause silent block loss).

Part 3 of 4: SSZ types → get_header → submit_block → validation bypassing

## Attribution

This work builds on the SSZ builder flow originally implemented by @eserilev 
in #252 and ported to the current codebase by @jclapis in #403.

Co-authored-by: Eitan Seri-Levi <eserilev@ucsc.edu>
Co-authored-by: Joe Clapis <jclapis@outlook.com>


crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -88,40 +126,42 @@ pub async fn submit_block<S: BuilderApiState>(
...
-            &request_api_version,
-            fork_name,
+            request_api_version,
         )
         .await
         {
             Ok(response) => {
-                // If the original request was for v2 but we had to fall back to v1, return a v2
-                // response
+                // If the original request was for v2 but we had to fall back to v1, the
```
0w3n-d: Would be great if these fallbacks could be removed. The v1 endpoint has been deprecated for a long time now. But relays still support it because boost clients still call it and boost clients still call it because it's available as a fallback.

I'd be surprised if there are still relevant relays that don't yet support v2 at this point.

JasonVranek: agreed it would be nice to prune the fallbacks but in the off chance the edge case is hit, this is purely more redundancy with no latency penalty. (I believe during kurtosis testing we hit this all the time too hence why it was added). I'd argue keep for now and after gloas v1/v2 are deprecated anyways

crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -191,90 +364,152 @@ async fn send_submit_block(
...
+            relay_id = relay.id.as_ref(),
+            status = %res.status(),
+            "relay rejected SSZ content-type, resubmitting block with JSON content-type"
+        );
+        res = match relay
+            .client
+            .post(url.as_ref().clone())
+            .timeout(Duration::from_millis(timeout_ms))
+            .headers(headers)
+            .body(serde_json::to_vec(&signed_blinded_block).unwrap())
```
ninaiiad: unwrap -> map_err into PbsError ?

crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -191,90 +364,152 @@ async fn send_submit_block(
...
+    if matches!(res.status(), StatusCode::NOT_ACCEPTABLE | StatusCode::UNSUPPORTED_MEDIA_TYPE,) {
+        warn!(
+            relay_id = relay.id.as_ref(),
+            status = %res.status(),
+            "relay rejected SSZ content-type, resubmitting block with JSON content-type"
+        );
+        res = match relay
+            .client
+            .post(url.as_ref().clone())
+            .timeout(Duration::from_millis(timeout_ms))
```
ninaiiad: because this doesn't run concurrently with the previous call, this function can take timeout_ms x2 and then the `submit_block_with_timeout`'s accounting of remaining_timeout_ms is incorrect:
```rust
remaining_timeout_ms = timeout_ms.saturating_sub(start_request.elapsed().as_millis() as u64);
```

crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -191,90 +364,152 @@ async fn send_submit_block(
...
-                fork_name,
-            )
+/// Decode an SSZ-encoded submit_block response
+fn decode_ssz_payload(
+    response_bytes: &[u8],
+    fork: ForkName,
+) -> Result<SubmitBlindedBlockResponse, PbsError> {
+    let data = PayloadAndBlobs::from_ssz_bytes_by_fork(response_bytes, fork).map_err(|e| {
+        PbsError::RelayResponse {
+            error_msg: (format!("error decoding relay payload: {e:?}")).to_string(),
```
ninaiiad: don't need .to_string()

crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -191,90 +364,152 @@ async fn send_submit_block(
...
+            status = %res.status(),
+            "relay rejected SSZ content-type, resubmitting block with JSON content-type"
+        );
+        res = match relay
+            .client
+            .post(url.as_ref().clone())
+            .timeout(Duration::from_millis(timeout_ms))
+            .headers(headers)
+            .body(serde_json::to_vec(&signed_blinded_block).unwrap())
+            .header(CONTENT_TYPE, EncodingType::Json.to_string())
```
ninaiiad: EncodingType::Json.content_type_header()

crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -160,22 +200,155 @@ async fn submit_block_with_timeout(
...
+    let mut res = match relay
         .client
-        .post(url)
+        .post(url.as_ref().clone())
         .timeout(Duration::from_millis(timeout_ms))
-        .headers(headers)
-        .json(&signed_blinded_block)
+        .headers(headers.clone())
+        .body(signed_blinded_block.as_ssz_bytes())
+        .header(CONTENT_TYPE, EncodingType::Ssz.to_string())
```
ninaiiad: EncodingType::Ssz.content_type_header()

crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -160,22 +200,155 @@ async fn submit_block_with_timeout(
...
+    timeout_ms: u64,
+    retry: u32,
+    api_version: BuilderApiVersion,
+) -> Result<Option<SubmitBlindedBlockResponse>, PbsError> {
+    // Send the request
+    let block_response = send_submit_block_impl(
+        relay,
+        url,
+        timeout_ms,
+        proposal_info.headers.clone(),
```
ninaiiad: let's pass by reference, we need to clone (not move) inside once anyway

ninaiiad (changes requested):

crates/pbs/src/mev_boost/submit_block.rs
0w3n-d: Yeah fair point about Gloas. Hardly worth trying to remove it now at this stage when gloas is not too far away.

ManuelBilbao (dismissed):

crates/pbs/src/mev_boost/submit_block.rs
```diff
@@ -160,22 +200,155 @@ async fn submit_block_with_timeout(
...
+            let expected_commitments = &blinded_block.body.blob_kzg_commitments;
+
+            validate_unblinded_block(
+                expected_block_hash,
+                got_block_hash,
+                expected_commitments,
+                &response.data.blobs_bundle,
+                response.version,
+            )
+        }
```
ninaiiad, line 236: do we need to support Electra?

crates/pbs/src/mev_boost/register_validator.rs
ninaiiad, line 58: This is a separate issue, but for a future reference it would be a bit cleaner to pass these by reference up until `send_register_validator` and only do the clone in the actual client request call

Right now we're cloning this for each relay in a loop here and then in the retry loop inside the `send_register_validator_with_timeout`

ninaiiad (approved):

crates/pbs/src/mev_boost/register_validator.rs
```diff
@@ -35,17 +35,19 @@ pub async fn register_validator<S: BuilderApiState>(
...
     send_headers.insert(USER_AGENT, get_user_agent_with_version(&req_headers)?);
 
-    // prepare the body in advance, ugly dyn
-    let bodies: Box<dyn Iterator<Item = (usize, Bytes)>> =
+    // prepare the body in advance
+    let bodies: Vec<(usize, Bytes)> =
         if let Some(batch_size) = state.config.pbs_config.validator_registration_batch_size {
-            Box::new(registrations.chunks(batch_size).map(|batch| {
-                // SAFETY: unwrap is ok because we're serializing a &[serde_json::Value]
-                let body = serde_json::to_vec(batch).unwrap();
```
ltitanb, line 43: why change this in this PR?

crates/pbs/src/mev_boost/register_validator.rs
```diff
@@ -35,17 +35,19 @@ pub async fn register_validator<S: BuilderApiState>(
         .insert(HEADER_START_TIME_UNIX_MS, HeaderValue::from_str(&utcnow_ms().to_string())?);
     send_headers.insert(USER_AGENT, get_user_agent_with_version(&req_headers)?);
 
-    // prepare the body in advance, ugly dyn
-    let bodies: Box<dyn Iterator<Item = (usize, Bytes)>> =
```
ltitanb, line 39: this was intentional to avoid the vec allocation, why change it?

crates/pbs/src/mev_boost/submit_block.rs
JasonVranek: You were right, removed the v2→v1 fallback since a v2 caller requires exactly 202 and never accepts a 200 body so the BN just fails regardless of commit-boost trying to add the redundancy

ltitanb (approved):

ManuelBilbao (approved):

PR #481, opened 2026-07-22
perf(pbs): always request SSZ from relays on get_header

JasonVranek:
PBS decodes and validates every relay bid and the route re-encodes the winning bid to the beacon node's Accept regardless, so the relay->PBS wire format never changes what the BN receives. Always request SSZ (JSON fallback) instead of mirroring the caller's format: SSZ is smaller on the wire and faster to decode, which matters across N relay bids on the proposal critical path. Mirrors the same change already made to submit_block.

### DIFF
--- a/crates/common/src/wire.rs
+++ b/crates/common/src/wire.rs
@@ -276,38 +276,6 @@ fn essence_encoding(mt: &MediaType) -> Option<EncodingType> {
     }
 }
 
-/// Compute the q-value for the `index`-th preferred encoding when building an
-/// outbound `Accept` header. The first entry gets q=1.0, each subsequent entry
-/// decreases by 0.1, and the value is clamped to a minimum of 0.1 so we never
-/// emit q=0 (which per RFC 7231 §5.3.1 means "not acceptable").
-fn accept_q_value_for_index(index: usize) -> f32 {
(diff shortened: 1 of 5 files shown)


crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -147,21 +147,18 @@ pub async fn get_header<S: BuilderApiState>(
...
+    // re-encodes the winning bid to the BN's Accept. So always request SSZ from
+    // the relay (smaller on the wire, faster to decode than JSON); JSON is the
+    // fallback for a relay that can't do SSZ. The BN's own format preference is
+    // applied later by the route, not here.
+    send_headers.insert(
+        ACCEPT,
+        build_outbound_accept(AcceptedEncodings {
+            primary: EncodingType::Ssz,
+            fallback: Some(EncodingType::Json),
         }),
```
ManuelBilbao: This can be `HeaderValue::from_static()` to avoid allocation. Even more, `build_outbound_accept` is only used twice with the same arguments, so we can remove it in favour of a constant

crates/common/src/wire.rs
ManuelBilbao: Note that this function is not used anymore

ManuelBilbao (approved):

ltitanb (approved):

PR #483, opened 2026-07-28
ws get_header bids stream draft

ninaiiad:


config.example.toml
```diff
@@ -92,6 +92,15 @@ headers = { X-MyCustomHeader = "MyCustomValue" }
 # GET parameters to add to each request URL for this relay
 # OPTIONAL
 get_params = { param1 = "value1", param2 = "value2" }
+# How to fetch headers from this relay.
```
ltitanb, line 95: I think a better config here is to just have a bool to say whether the relay supports streaming headers, the endpoint should be deterministic like `/eth/v1/builder/header_stream`

crates/common/src/pbs/relay.rs
```diff
@@ -120,6 +156,35 @@ impl RelayClient {
...
+        Ok(match &self.config.get_header {
+            GetHeaderTransport::Http => {
+                GetHeaderRequest::Http(self.get_header_url(slot, parent_hash, validator_pubkey)?)
+            }
+            GetHeaderTransport::Stream(base) => {
+                let mut url = base.clone();
+                // Push segments rather than Url::join: join drops both the last
+                // segment of a base without a trailing slash and its query
+                url.path_segments_mut()
+                    .map_err(|_| PbsError::WebSocket("stream url cannot be a base".to_string()))?
```
ltitanb: any error of this sort should be handled at startup, if it is already we should unwrap here

JasonVranek:
I like the simplicity of per-slot streams vs the old PoC establishing long lived connections. Q is if handshake overhead is neglible or will be problematic / is this enough time for the relay to establish reliable RTTs?

crates/pbs/src/mev_boost/get_header.rs
```diff
@@ -249,15 +250,31 @@ async fn fetch_parent_block(
...
     request_info: Arc<RequestInfo>,
     relay: RelayClient,
     ms_into_slot: u64,
-    mut timeout_left_ms: u64,
+    timeout_left_ms: u64,
 ) -> Result<Option<GetHeaderResponse>, PbsError> {
     let params = &request_info.params;
-    let url = relay.get_header_url(params.slot, &params.parent_hash, &params.pubkey)?;
 
+    match relay.get_header_request(params.slot, &params.parent_hash, &params.pubkey)? {
```
JasonVranek, line 252: get_header_from_relay forces WS or HTTP instead of firing them off concurrently which might be good for redundancy, esp as WS rollouts are new. You could always duplicate the relay in your config but then you send duplicates reg + status + submit_block and have to more to manage in configs. More of an optimization / UX improvement but can consider firing concurrent HTTP/WS.

crates/pbs/src/mev_boost/get_header_ws.rs
```diff
@@ -0,0 +1,320 @@
...
+    let data = SignedBuilderBid::from_ssz_bytes_by_fork(&bid_bytes, fork).map_err(|err| {
+        PbsError::SSZDecode {
+            err: format!("error decoding relay payload from ws stream: {err:?}"),
+            fork,
+        }
+    })?;
+    let response = GetHeaderResponse { version: fork, data, metadata: Default::default() };
+
+    let start_validate = Instant::now();
+    let validated = validate_get_header_response(request_info, relay, &response);
```
JasonVranek, line 198: Keep the latest ≠ keep the highest-value, is this intentional? If the relay happens to return a higher bid earlier in the stream it would be overwritten. 

Related, applying `validate_get_header_response` only at the last header means a valid header could have been discarded while the latest is invalid. Potentially can maintain a buffer of N frames. Then either pay N*validation cost at the end or as an optimization validate as frames fill the buffer off the hot path.

crates/pbs/src/mev_boost/get_header_ws.rs
```diff
@@ -0,0 +1,320 @@
...
+    };
+
+    if let Some(first_bid_latency) = first_bid_latency {
+        RELAY_LATENCY
+            .with_label_values(&[GET_HEADER_ENDPOINT_TAG, &relay.id])
+            .observe(first_bid_latency.as_secs_f64());
+    }
+    record_status("200", relay);
+
+    let data = SignedBuilderBid::from_ssz_bytes_by_fork(&bid_bytes, fork).map_err(|err| {
```
JasonVranek: can reuse decode_json_payload

crates/pbs/src/mev_boost/get_header_ws.rs
```diff
@@ -0,0 +1,320 @@
...
+        record_status("204", relay);
+        return Ok(None);
+    };
+
+    if let Some(first_bid_latency) = first_bid_latency {
+        RELAY_LATENCY
+            .with_label_values(&[GET_HEADER_ENDPOINT_TAG, &relay.id])
+            .observe(first_bid_latency.as_secs_f64());
+    }
+    record_status("200", relay);
```
JasonVranek: success recorded before the bid is decoded/validated

crates/pbs/src/mev_boost/get_header_ws.rs
```diff
@@ -0,0 +1,320 @@
...
+    let connect = connect_async_tls_with_config(
+        request,
+        Some(config),
+        true,
+        Some(Connector::Rustls(tls_config().clone())),
+    );
+    let (mut stream, _) = match timeout_at(deadline, connect).await {
+        Ok(Ok(connected)) => connected,
+        Ok(Err(err)) => {
+            record_status(TIMEOUT_ERROR_CODE_STR, relay);
```
JasonVranek: `TIMEOUT_ERROR_CODE_STR` not timeout for this one

crates/pbs/src/mev_boost/get_header_ws.rs
```diff
@@ -0,0 +1,320 @@
...
+                }
+            }
+        }
+    }
+
+    drop(stream);
+
+    let Some((fork, bid_bytes)) = latest else {
+        if let Some(err) = stream_error {
+            record_status(TIMEOUT_ERROR_CODE_STR, relay);
```
JasonVranek: also not timeout

crates/pbs/src/mev_boost/get_header_ws.rs
vladimir-ea: I believe it is intentional - bids can decline in value due to transaction cancellations which builders & relays honour.

crates/pbs/src/mev_boost/get_header.rs
ninaiiad: it's a design choice for the first draft to just do WS

also I think it's cleaner to duplicate the relay in the config to dispatch both HTTP & WS

config.example.toml
```diff
@@ -92,6 +92,18 @@ headers = { X-MyCustomHeader = "MyCustomValue" }
...
+#   "http"     -> one request per get_header, at the relay url above
+#   "stream"   -> websocket stream of bid updates, only for relays that support it. Connects to
+#                 ws(s)://<relay url>/eth/v1/builder/header_stream/{slot}/{parent_hash}/{pubkey}.
+#                 The stream stays open for the whole request window and the last bid pushed is the
+#                 one used.
+# OPTIONAL, DEFAULT: "http"
+get_header = "http"
+# Name of the environment variable holding this relay's API key. It is sent as `X-Api-Key` when opening a get_header stream, for relays that require one.
+# Only used when `get_header = "stream"`.
+# OPTIONAL
```
ltitanb: do we need this? we already have customizable headers and params above for get header which operators use for authentication with relays

crates/common/src/pbs/relay.rs
```diff
@@ -75,31 +111,64 @@ impl RelayClient {
             }
         }
 
+        let mut stream_headers = match config.get_header {
```
ltitanb: are we still including the custom headers?
```toml
# Headers to send with each request for this relay
# OPTIONAL
headers = { X-MyCustomHeader = "MyCustomValue" }
```

ltitanb (dismissed):

crates/pbs/src/mev_boost/get_header_ws.rs
```diff
@@ -0,0 +1,327 @@
...
+        1 => ForkName::Altair,
+        2 => ForkName::Bellatrix,
+        3 => ForkName::Capella,
+        4 => ForkName::Deneb,
+        5 => ForkName::Electra,
+        6 => ForkName::Fulu,
+        7 => ForkName::Gloas,
+        _ => return None,
+    })
+}
```
ltitanb, line 71: here seems super easy that in the future we add a fork but forget to add it here since we are not checking that we have all the variants

crates/common/src/pbs/relay.rs
ninaiiad: yes

crates/pbs/src/mev_boost/get_header_ws.rs
ninaiiad: added a test for this

ltitanb (dismissed):

crates/pbs/src/mev_boost/get_header_ws.rs
```diff
@@ -0,0 +1,341 @@
...
+        ?validate_latency,
+        version = ?fork,
+        value_eth = format_ether(*response.value()),
+        block_hash = %response.block_hash(),
+        updates,
+        invalid_frames,
+        "received new header from ws stream"
+    );
+
+    validated?;
```
JasonVranek: here and above at `decode_ssz_payload(&bid_bytes, fork)?;` we could exit early which could drop the `info!` and possibly `record_status` which may leave gaps when debugging. Worse is `info!` fires -> validation fails, since to a user the `info` looks like a success.

maybe can consider calling `record_status` only on the return result of `get_header_ws`

crates/pbs/src/mev_boost/get_header_ws.rs
```diff
@@ -0,0 +1,341 @@
...
+    let connect = connect_async_tls_with_config(
+        request,
+        Some(config),
+        true,
+        Some(Connector::Rustls(tls_config().clone())),
+    );
+    let (mut stream, _) = match timeout_at(deadline, connect).await {
+        Ok(Ok(connected)) => connected,
+        Ok(Err(err)) => {
+            let rejected = match &err {
```
JasonVranek: consider also checking `res.body()` so the reason for the error is surfaced not just the status

crates/pbs/src/mev_boost/get_header_ws.rs
```diff
@@ -0,0 +1,341 @@
...
+    RELAY_STATUS_CODE.with_label_values(&[code, GET_HEADER_ENDPOINT_TAG, &relay.id]).inc();
+}
+
+/// One TLS config for every stream connection. Left to tokio-tungstenite it is
+/// rebuilt per connect, which reparses the root store and, worse, gives each
+/// connection its own session cache: every slot then pays a full handshake
+/// instead of a resumed one. The provider is named explicitly because rustls is
+/// built with both `ring` and `aws-lc-rs` here, so the default builder needs a
+/// process-wide install to pick one.
+fn tls_config() -> &'static Arc<ClientConfig> {
```
JasonVranek, line 291: in `signer/src/service.rs` we have `install_crypto_provider()`. can consider moving to `cb_common` and reusing in both pbs/signer. then don't have to worry about a runtime panic (avoid issue we had from #480)

ltitanb (dismissed):

crates/pbs/src/mev_boost/get_header_ws.rs
gd-0, line 1: If this fails do we fall back to http?

ninaiiad: added this

JasonVranek (dismissed):

ltitanb (approved):

JasonVranek (approved):

PR #488, opened 2026-08-11
ci(docs): manual, main-only docs deploy on demand

JasonVranek:
Replace the abandoned `stable`-branch deploy (frozen at v0.9.3) with a Pages deploy that publishes the docs on the `main` branch.

The deploy is manual (`workflow_dispatch`), not push-triggered which has caused issues in the past.

test-docs runs the build on any PR that touches the docs or their linked inputs, so breakage is caught before merge.


.github/workflows/docs.yml
```diff
@@ -1,27 +1,46 @@
 name: Docs
 
+# Manual only: click "Run workflow" to publish the docs currently on the
+# selected branch (defaults to the repo default branch). Publishing is a
+# deliberate act, not a side effect of a merge, and a failed run can simply be
+# re-run without pushing a new commit.
```
ManuelBilbao: This contradicts the next comment: the workflow only runs if the selected branch is main. Also, I'd cut the second sentence, it's redundant, and reads like LLM verbosity

ManuelBilbao (commented):
Eventually we should move the building job to custom action so `test-docs` and `docs` use the same build script

ManuelBilbao (approved):

ltitanb (approved):

PR #498, opened 2026-09-11
feat(pbs): load relay headers from a secret file or env var

JasonVranek:
## Summary

A relay's custom headers, which is how a relay `X-Api-key` is supplied, can now come from a secret file or an environment variable instead of the config file:
```toml
headers = { X-Api-Key = { file = "/run/secrets/relay-key" } }
# or: headers = { X-Api-Key = { env = "RELAY_API_KEY" } }
# or: headers = { X-Api-Key = "literal" }   # unchanged
```

The value is read when the relay client is built, at startup and on every config reload, so a rotated secret is picked up by a reload.

Header values are marked sensitive so they never appear in debug output, and a literal never appears in the config's `Debug`. A header loaded from a secret logs its source and a 4-byte fingerprint of the value, never the value, so a rotation is visible across reloads.

`commit-boost init` mounts every `file` path read-only into the PBS container at the same path and passes every `env` variable through from the compose environment, for default and mux relays alike. A relative or missing file fails `init` rather than producing a compose file docker fills with a root-owned directory.


crates/cli/src/docker_init.rs
```diff
@@ -269,6 +269,30 @@ fn create_pbs_service(service_config: &mut ServiceCreationInfo) -> eyre::Result<
         }
     }
 
+    // Relay header secret files, mounted read-only at their own path so the
+    // config's `{ file = ... }` resolves inside the container unchanged
+    for path in cb_config.relay_header_files() {
+        eyre::ensure!(
+            path.is_absolute(),
```
ManuelBilbao, line 276: Why do we need this to be absolute?

JasonVranek: relative made it much more complex, something we can consider later but would say as a follow on PR

ninaiiad (approved):

ltitanb (approved):

ManuelBilbao (approved):

